PR #30, opened 2013-07-27
ASN.1 basic format types/parsers/encoders

oh2mqk:
Just one directory set of javadoc patches.

### DIFF
--- a/core/src/main/java/org/bouncycastle/asn1/ASN1ApplicationSpecificParser.java
+++ b/core/src/main/java/org/bouncycastle/asn1/ASN1ApplicationSpecificParser.java
@@ -2,9 +2,16 @@ package org.bouncycastle.asn1;
 
 import java.io.IOException;
 
+/**
+ * Interface for parse of ApplicationSpecific objects.
+ */
 public interface ASN1ApplicationSpecificParser
     extends ASN1Encodable, InMemoryRepresentable
 {
+    /**
+     * Read an object from the input stream.
+     * @throws IOException for bad input stream.
+     */
     ASN1Encodable readObject()
         throws IOException;
 }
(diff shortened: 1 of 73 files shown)


RajeshBhatwar:
This is my request can you pleas demonstrate a simple example to write a Encode / Decode method in java using ASN.1 syntax.

For example i have valid ASN.1 syntax of some parser just say for BER parser so i have two option 
1) I can use any ASN.1 compiler (Most of the compilers are paid and hardly we can find open source on top of that most of the generated java files will contain errors because of limited support for latest syntax Or for  some other reason ) to generate .java files by using ASN.1 syntax and get a Encode / Decode method. It is more easy way to start with ASN.1 compilers, but the ASN.1 compilers are so expansive most of the new bees cannot offer (its around $16000 / 7000 Euros).

2) To avoid the expanse we should write our own JAVA code by referring ASN.1 syntax. As a starter / new to ASN.1 / for a new bees it is very tough task (this is my feeling) because i don't find any reference (after spending allmost 6+ days) / guidelines / any white papers / samples to write ASN.1 syntax to Java classes  using some APIs like Bouncy Castle.


So, after seeing your Github project / repository i personally feel you can write a simple project which demonstrate by considering simple ASN.1 syntax to create JAVA encode / Decode methods with setter ,getter ,constructors and so on...

 It is a request and it will help for lot more peoples like me ....

If you want to tell something then you can write a mail @ rajeshhatwar2027@gmail.com 

Thanks for the need full help.

oh2mqk:
Sorry, no help here for compiling parsers from ASN.1 syntax definition.
The BouncyCastle parsers are manually coded.

In my opinnion they are not that difficult to understand, once you spot which syntax fragment is being parsed at which part.

You were now looking at primitive types, I suggest you look at more complex types to see parser examples, for example: org.bouncycastle.asn1.x509.Certificate getInstance() methods do.

RajeshBhatwar:
Thanks for your suggestion i will look into that

bcgit:
The last of these files have finally been reviewed an incorporated.

PR #76, opened 2014-07-02
BCrypt implementation

timw:
Implementation of raw and OpenBSD/crypt style bcrypt password hash functions.

Raw and OpenBSD style are separated.
The OpenBSD/crypt style is most useful for compatibility with existing uses or clients wanting a simple hash+params encoding.
The raw style could be plugged into a future password hash API quite easily (i.e. based on the PHC API).

To reuse the existing Blowfish implementation (having an unverified copy is the cause of at least one major bug in crypt_blowfish) BlowfishEngine is extended slightly with hooks to allow the key schedule to be overridden and for salt to be mixed in during table processing in key setup. Performance impact should be negligible given the existing cost of the Blowfish key setup.

In the OpenBSD implementation the 2 (original), 2a, 2b and 2y formats are supported (a/b/y are algorithmically equivalent, but represent bug fixes in various crypt implementations). The crypt_blowfish 2x and 'safe 2a' modes are not implemented since they require borking the BlowfishEngine (where the bug was identified).

Cross tested with JBCrypt and OpenWall crypt_blowfish test vectors (links in unit tests).

Also added checking for defined key sizes in BlowfishEngine, with a constructor/factory for uses needing unrestricted key lengths (bcrypt being one of them). That's in a separate commit, so feel free to drop/amend.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/engines/BlowfishEngine.java
+++ b/core/src/main/java/org/bouncycastle/crypto/engines/BlowfishEngine.java
@@ -7,15 +7,70 @@ import org.bouncycastle.crypto.OutputLengthException;
 import org.bouncycastle.crypto.params.KeyParameter;
 
 /**
- * A class that provides Blowfish key encryption operations,
- * such as encoding data and generating keys.
- * All the algorithms herein are from Applied Cryptography
- * and implement a simplified cryptography interface.
+ * Implementation of the <a href="https://www.schneier.com/blowfish.html">Blowfish block cipher</a>
+ * as described in Applied Cryptography.
  */
-public final class BlowfishEngine
+public class BlowfishEngine
 implements BlockCipher
 {
-    private final static int[] 
+
+    /**
+     * A cyclic producer of 4 byte words.
+     */
+    protected interface WordCycle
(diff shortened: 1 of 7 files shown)


bcgit:
Finally added with some modifications. Thanks.

timw:
Meh. I tried to avoid copying the entire Blowfish implementation just to modify the key setup tweak (200 loc vs 600 loc).

There's at least one bug in the BCrypt as committed, amusingly one [known in jBCrypt since 2013](https://bugzilla.mindrot.org/show_bug.cgi?id=2097) but actually fixed only 2 days ago.

bcgit:
Might be worth fixing it now then! It looks straightforward enough... is there a test vector for it?

Copied the code to avoid making BlowfishEngine non-final... There's probably another way but it will do for now. It's a pity it wound up in generators.

timw:
Yeah, there's the odd time I hanker for a C++ friend equivalent. 
I don't think there's a test vector (cost == 31 is a little OTT for unit tests) - it's probably sufficient to fix with a comment given it's an edge case.

bcgit:
cost 31 now fixed.

PR #82, opened 2014-07-25
PGP: Add constants for GNU protection modes in PGP S2K class

dschuermann:
These are the two actually used protection modes in GnuPG. I found them by reading the gpg sourcecode.


bcgit:
Added. Thanks!

PR #83, opened 2014-07-25
Stop TLS testcases causing failure on JUnit constructor requirements

rtyley:
If you run the Gradle build, you'd see TlsTestCase & DTLSTestCase cause a build failure:

```
junit.framework.AssertionFailedError:
Class org.bouncycastle.crypto.tls.test.DTLSTestCase has no public
constructor TestCase(String name) or TestCase()
```

The reason is that both these testcases actually have two-arg constructors (because they need a bunch of extra configuration data), so don't fit the normal no-arg or one-arg constructor requirement of classes extending `TestCase`.

The fix used here is to employ JUnit 'parameterized tests', rather than a TestSuite:

https://github.com/junit-team/junit/wiki/Parameterized-tests

P.S. You can see also see the Gradle build failure in Travis CI here:

https://travis-ci.org/rtyley/spongycastle/builds/29981411#L1052-L1056

(this is why BC needs https://github.com/bcgit/bc-java/pull/80 !)


peterdettman:
Fixed in an alternative, but straight-forward way. Parameterized tests appear to depend on annotations, which are of course only Java 1.5+. Instead, we just pass a dummy test if we weren't created by the test suite.

PR #86, opened 2014-08-14
PGP: Dont write version header by default to armored messages

dschuermann:
By telling which version of Bouncy Castle has been used in armored outputs, it is much easier to find exported keys on the Internet which have been generated by a potentially broken version of the library.
Writing a version header should be totally optional.
see also http://blog.cryptographyengineering.com/2014/08/whats-matter-with-pgp.html
and https://mailman.boum.org/pipermail/tails-dev/2013-August/003435.html

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredOutputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredOutputStream.java
@@ -93,8 +93,6 @@ public class ArmoredOutputStream
     String          footerStart = "-----END PGP ";
     String          footerTail = "-----";
 
-    String          version = "BCPG v@RELEASE_NAME@";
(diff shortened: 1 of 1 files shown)


bcgit:
I've modified the code so the version is now optional - default behaviour is preserved though. I see the argument, but this is one of those "dammed if you do, dammed if you don't" kind of things. With the new code

ArmoredOutputStream.setHeader(ArmoredOutputStream.VERSION_HDR, null);

will clear the version header.

dschuermann:
Thanks for fixing this. BTW: PGP version header can be really problematic: http://www.networkworld.com/article/2904395/microsoft-subnet/mistakes-that-betrayed-anonymity-of-former-dea-agent-and-silk-road-investigator.html

PR #88, opened 2014-08-20
Allow user to specify an alternate iterationCount

dstutz:
Iteration count is hard-coded to 1024 in this file.  The change allows user to specify an alternate value to use (preferably higher).

Alternative implementations would be adding an additional constructor or changing parameters but I didn't know which method you'd prefer so I went with something that's not very intrusive by adding a method.


bcgit:
Added. Thanks.

PR #89, opened 2014-08-23
PGP: Expose S2K packet

dschuermann:
To know if we deal with a key that contains no secret key (GNU extension) we would like to expose the S2K packet. This pull requests adds a method to PGPSecretKey to get the packet.


bcgit:
Added

PR #91, opened 2014-08-27
some small improvements when searching for issuer certificates

Vampire:


bcgit:
Applied with some modifications.

PR #96, opened 2014-10-17
Add non-blocking API to TlsProtocol

rdicroce:
This PR supersedes #68. It's not a split of that PR, it's a complete rewrite that has a cleaner API and less invasive changes. The TLS-SRP changes will be done as part of a separate PR, as requested.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/tls/ByteQueueInputStream.java
+++ b/core/src/main/java/org/bouncycastle/crypto/tls/ByteQueueInputStream.java
@@ -1,0 +1,70 @@
+package org.bouncycastle.crypto.tls;
+
+import java.io.InputStream;
+
+public class ByteQueueInputStream extends InputStream
+{
+
+    private ByteQueue buffer;
+    
+    public ByteQueueInputStream()
+    {
+        buffer = new ByteQueue();
+    }
+    
+    public void addBytes(byte[] bytes)
+    {
+        buffer.addData(bytes, 0, bytes.length);
+    }
+    
+    public int peek(byte[] buf)
+    {
+        int bytesToRead = Math.min(buffer.available(), buf.length);
+        buffer.read(buf, 0, bytesToRead, 0);
+        return bytesToRead;
+    }
+    
+    @Override
+    public int read()
+    {
+        if (buffer.available() == 0)
+        {
+            return -1;
+        }
+        return buffer.removeData(1, 0)[0] & 0xFF;
+    }
+
+    @Override
+    public int read(byte[] b)
+    {
+        return read(b, 0, b.length);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len)
+    {
+        int bytesToRead = Math.min(buffer.available(), len);
+        buffer.removeData(b, off, bytesToRead, 0);
+        return bytesToRead;
+    }
+
+    @Override
+    public long skip(long n)
+    {
+        int bytesToRemove = Math.min((int) n, buffer.available());
+        buffer.removeData(bytesToRemove);
+        return bytesToRemove;
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


timw:
This is another of those changes where a 1.5/6/7/8+ source tree + build to newer JVM features could be useful (java.nio dependencies clash with the desire to keep BC building/running on JDK 1.1ish).
It looks like this PR could be modified to remove the java.nio deps though...

rdicroce:
I'm not sure what the rationale is for continuing to support JDKs < 5. As far as I know, Oracle doesn't offer any kind of support for those versions. 5 and 6 are no longer receiving public updates, and 7 will be joining them in a couple months. The only other JVM that I've heard of people using is IBM's; it looks like they're still supporting 1.4.

Bottom line: if you're using an older than dirt JVM, you should really consider upgrading. A lot of other Java libraries are ditching support for older versions of Java anyway. 6 seems to have become the baseline, and some even require 7 (Apache Lucene comes to mind).

timw:
This topic pops up semi-regularly on the BC dev list, but tl;dr: BC is used a lot in embedded systems, J2ME etc. that are fairly frozen in time wrt Java version, so portability is valued.

peterdettman:
We're going to defer this to the next release again, sorry (currently putting 1.52 together). I'm sympathetic to the JDK constraints, believe me. As @timw alludes, a path forward is to externalize anything JDK (or IO-lib) dependent via an API, and have a separate module/project supply an implementation.

rdicroce:
Creating an API might be a good idea anyway, now that I'm thinking about it. It could simplify integration with other libraries like Netty, which has its own ByteBuf class. I might explore that if I have some time.

rdicroce:
I've changed the PR so it no longer uses ByteBuffer, only byte arrays now. Also rebased it on top of current master. And added some additional testing, which revealed the original PR had a serious bug when partial records arrived. That issue is fixed now.

mezz:
Any news now that 1.52 is released?

rivervilla:
Any chance non-blocking TLS and/or DTLS will be in 1.53?

peterdettman:
Honestly, I had missed the fact that @rdicroce had updated this PR. My apologies for that. We are busy with the FIPS work just now, but I will find some time to review this and see if we can get it into 1.53.

ivelin:
@peterdettman Can you please share an update on timeframe for NIO?

peterdettman:
Merged. Thanks @rdicroce for the patch, particularly for reworking it to accommodate our Java version constraints.

rdicroce:
@peterdettman It looks like this did not make it into 1.53, so I assume it will be in 1.54 then? Also it looks like the merge has not been pushed to GitHub yet.

rivervilla:
I'll look at changes but this is true non-blocking meaning partial records will not cause problems. I'm also in need of a negotiation callback so that user may allow client cert privacy through rehandshake after connection is protected.  Yes different issue.  Thanks.

peterdettman:
No, it didn't make it into 1.53. It'll be in the next beta build of 1.54 which shouldn't be too long. It also takes a little while for our git commits to mirror back to github (maybe just check back tomorrow).

bcgit:
This is now up on http://www.bouncycastle.org/betas (154b03 or later). The code changes will appear in the main branch here shortly - the github mirror has been updated, it just needs to sync up.

sscheck:
What are the relevant classes for non-blocking mode? I don't see any TlsServer or DatagramTransport implementations that obviously indicate non-blocking operation.

rdicroce:
For TLS, the relevant methods are on TlsServer/ClientProtocol. I didn't do any work on DTLS, and AFAICT the release notes for 1.54 are incorrect in saying that a non-blocking API was added for DTLS.

PR #97, opened 2014-10-17
Implement server side of TLS-SRP

rdicroce:
This PR supersedes #68. Mostly the same code as before, just as a separate PR. As noted in #68:

I'm not sure if I've satisfied the requirements of the RFC completely, but at the very least, the test I added for TLS-SRP is passing. Note that this pull request implements support for SRP cipher suites only; SRP-RSA and SRP-DSS cipher suites are not supported. (More accurately, I have not tested them, so they may or may not work.)


peterdettman:
This has now been implemented. Not merged, but heavily derived from this PR, with extensions to support server-side SRP_DSS and SRP_RSA, and an implementation of "unknown identity simulation".

Thanks for the patch!

PR #101, opened 2014-12-22
Ecies options - allow EC point compression for ECIES scheme

halfmanhalffish:
This change allows clients to specify EC Point Compression for the output of the ECIES encryption routine.  This reduces output by 32 bytes (for EC key size 256).  Clients specify point compression by using a newly added constructor for org.spongycastle.jce.spec.IESParameterSpec in which the last argument is a boolean flag 'usePointCompression'.

The default remains (as before) no point compression.
This change does not introduce any API incompatibilities or changes.
Point Compression works equally with streaming (xor) mode or conventional block cipher.

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jce/spec/IESParameterSpec.java
+++ b/prov/src/main/java/org/bouncycastle/jce/spec/IESParameterSpec.java
@@ -14,6 +14,7 @@ public class IESParameterSpec
     private byte[] encoding;
     private int macKeySize;
     private int cipherKeySize;
+    private boolean usePointCompression = false;
     private byte[] nonce;
 
 
@@ -47,7 +48,7 @@ public class IESParameterSpec
         int macKeySize,
         int cipherKeySize)
     {
-        this(derivation, encoding, macKeySize, cipherKeySize, null);
+        this(derivation, encoding, macKeySize, cipherKeySize, null, false);
     }
 
     /**
@@ -65,6 +66,27 @@ public class IESParameterSpec
         int macKeySize,
         int cipherKeySize,
         byte[] nonce)
+    {
+        this(derivation, encoding, macKeySize, cipherKeySize, null, false);
+    }
+
+    /**
+     * Set the IES engine parameters.
+     *
+     * @param derivation    the optional derivation vector for the KDF.
+     * @param encoding      the optional encoding vector for the KDF.
+     * @param macKeySize    the key size (in bits) for the MAC.
+     * @param cipherKeySize the key size (in bits) for the block cipher.
+     * @param nonce         an IV to use initialising the block cipher.
+     * @param usePointCompression whether to use EC point compression or not (false by default)
+     */
+    public IESParameterSpec(
+        byte[] derivation,
+        byte[] encoding,
+        int macKeySize,
+        int cipherKeySize,
+        byte[] nonce,
+        boolean usePointCompression)
     {
         if (derivation != null)
         {
@@ -89,6 +111,7 @@ public class IESParameterSpec
         this.macKeySize = macKeySize;
         this.cipherKeySize = cipherKeySize;
         this.nonce = Arrays.clone(nonce);
+        this.usePointCompression = usePointCompression;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


prov/src/main/java/org/bouncycastle/jce/spec/IESParameterSpec.java
```diff
@@ -66,6 +67,27 @@ public IESParameterSpec(
         int cipherKeySize,
         byte[] nonce)
     {
+        this(derivation, encoding, macKeySize, cipherKeySize, null, false);
```
peterdettman: Note that 'null' should be 'nonce' here

peterdettman:
The possibility of using compressed points with ECIES is already supported in the current lightweight API, via KeyEncoder/KeyParser interfaces. I've modified ECIESTest to cover that usage. If I am understanding the IESEngine changes correctly, this PR uses the uncompressed encoding unconditionally as input to the KDF instead of the actually-used encoding. That would be a backward-compatibility break at this stage. In any case, what is or isn't input to the KDF seems to be a matter of configuration/convention, and the current way is simpler and no-less-correct to my knowledge.

Are you trying to be compatible with another library? It's still possible to include this behaviour, but it would have to be via (yet another) configuration option, or externalising the building of the KDF input somehow. The current approach will have to stay as the default.

I have applied your IESParameterSpec changes to enable point compression via the provider.

halfmanhalffish:
Hi Peter,

Nice to meet you and thanks for picking this up.

The null instead of nonce certainly looks like a careless mistake on my part, so thanks for spotting that.

As I understand it, point compression is only relevant to wire encoding and not to any other part of the process, including KDF input.  Also, again as I understand it, KDF input is security-sensitive.  Passing compressed points to the KDF would be a security cock-up.  I'm 99% confident of that.
If the above is correct, then it follows that upon decryption compressed points would have to be 'reflated' in order that the KDF input be the same as it was for encryption.  Without that, the algorithm would not be sound (ie decrypt what it encrypted).
Summarising my point: the existing KDF input is the uncompressed points, which is correct, and my Pull Request doesn't change that, because to use compressed points would be insecure.

You ask why I'm making this change - it's because my application is extraordinarily bandwidth sensitive and I want to save every byte I can.

Thanks
Philip

peterdettman:
I don't think there are any security issues in what type of encoding is sent to the KDF. Both compressed and uncompressed encodings contain exactly the same information after all. Indeed it is sometimes suggested to just use the x coordinate as KDF input, or even to hash the point encoding (or x coordinate only) before sending it to the KDF (see e.g. http://www.researchgate.net/publication/255970113_A_Survey_of_the_Elliptic_Curve_Integrated_Encryption_Scheme section 6).

In specs (e.g. http://www.secg.org/sec1-v2.pdf section 5.1), the KDF input is described in quite vague terms, the important thing obviously being that both ends agree on it. There's some additional discussion on KDFs in http://csrc.nist.gov/publications/nistpubs/800-56B/sp800-56B.pdf section 5.9.

So, if you are in control of both ends of your application, I see no reason not to use the current BC implementation, specifying point compression via the KeyEncoder implementation (see ECIESTest.doEphemeralTest for a working example). If we start to see users that need precise control over the KDF, we'll have to expand the API somehow to accommodate that, but we won't be changing the current default behaviour, to preserve backward-compatibility.

halfmanhalffish:
Hi again Peter,

I had not realised that point compression was possible using this lightweight API (which I assume means a non-Java Security API, presumably using just the core jar and not the 'prov' jar).  I tried, incidentally, to use the non-Provider/lightweight API but found that I could not get rid of the 'prov' jar dependency.  That's how I ended up using the provider API, and that's why I was operating on the assumption that the KDF was always given uncompressed points, never compressed ones.  Given that assumption, it would have been reckless for me to change the input to the KDF.

I think there is some value in general in being conservative (ie in consistently using one type of points, either compressed or uncompressed) if you take the view that complexity is the enemy of security.  By which I mean that the more code paths there are, the greater the attack surface and the more points of analysis that the good guys have to be certain about.

Thanks
Philip

PR #103, opened 2015-01-27
Fix encoding of EC private keys.

nikosn:
If the byte length of the key is shorter than the byte length of the order, the resulting octet string of the private key bits is also shortended. This does not match the specification in  RFC 5915 which demands:

      privateKey is the private key.  It is an octet string of length
      ceiling (log2(n)/8) (where n is the order of the curve) obtained
      from the unsigned integer via the Integer-to-Octet-String-
      Primitive (I2OSP) defined in [RFC3447].

OpenSSL has the same issue, https://rt.openssl.org/Ticket/Display.html?id=3644


bcgit:
Merged with some changes.

PR #104, opened 2015-01-29
Allow user to specify a different iteration count

dstutz:
Provide facility for user to change the iteration count from the hard-coded default of 1024.  I previously did a change for the MAC builder so this brings the same functionality to the encryption side.  This PR is for the JCE piece, I'll do another one for the BC side.


bcgit:
Merged. Thanks.

PR #105, opened 2015-01-29
Allow user to specify a different iteration count

dstutz:
Provide facility for user to change the iteration count from the hard-coded default of 1024.  This is the BC half of PR #104.


bcgit:
Applied. Thanks.

PR #107, opened 2015-02-10
Update CMSSignedData.java

lartiguePierre:
[ERROR FIX] error fixed in the couter signature verification of a CSMSignedData


bcgit:
Merged. Thanks. Faulty test fixed as well.

PR #108, opened 2015-02-11
ChaCha20/Poly1305 AEAD and 96 bit IV

timw:
Implementation of the ChaCha20/Poly1305 AEAD construction from draft-irtf-cfrg-chacha20-poly1305.
As part of the above, addition of 96 bit IV/32 bit counter split in ChaChaEngine.

A couple of sub-optimal bits:
- AEADBlockCipher is slightly abused here. This is an AEADCipher, but BC doesn't have that concept. If there's interest in improving the API structure here (especially since many the CAESER competition entries are similarly not block cipher modes) I'm happy to propose some changes to move that along.
- 96/32 split is only implemented for ChaChaEngine (not Salsa20) since it's a little unusual (@agl has commented that it's targeted at IPSec).

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/engines/ChaChaEngine.java
+++ b/core/src/main/java/org/bouncycastle/crypto/engines/ChaChaEngine.java
@@ -4,9 +4,22 @@ import org.bouncycastle.util.Pack;
 
 /**
  * Implementation of Daniel J. Bernstein's ChaCha stream cipher.
+ * <p>
+ * ChaCha uses a 128 or 256 bit key and a 64 bit nonce.
+ * <p>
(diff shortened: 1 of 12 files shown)


timw:
If this passes muster, I plan to rework the ChaCha/Poly1305 in the TLS package to the scheme in draft-mavrogiannopoulos-chacha-tls (which builds on this and supercedes draft-agl-tls-chacha20poly1305) - it'll have to wait for a codepoint assignment though.

peterdettman:
@timw We actually did a fair bit of work related to ChaCha and Poly1305 recently (now supporting draft-ietf-tls-chacha20-poly1305-04), and I overlooked the existence of this PR at the time, so I'm not quite sure how things stand here.

peterdettman:
Pretty sure this is obsolete now.

PR #109, opened 2015-02-12
AddCounterSigners bug fix in SignerInformation

lartiguePierre:
The old method was functionnal for only one counter signer. It added only one CounterSigner attribute for all counterSigners present in the input store. It adds one for each in this version.


bcgit:
Verification code for counter signatures is now recursive.

PR #111, opened 2015-03-05
Fixes DTLS negotiation when downgrading from 1.2 to 1.0.

gpolitis:
Hi, this PR fixes/hacks around the issue I've described here : https://www.bouncycastle.org/devmailarchive/msg14087.html


peterdettman:
I agree there is a problem with DTLS version negotiation. The RFCs are IMO ambiguous (particularly in relation to the record layer versions), but I guess we'll just try and figure out the rules that openssl and gnutls are using and take that as standard.

peterdettman:
We've now fixed DTLS version negotiation and the record-layer version handling (there are test cases covering your specific case of a client offering 1.2 connecting to a server with max 1.0). Thank you for raising the issue.

PR #114, opened 2015-06-03
Update CMSSignedDataParser.java

bschuette:
I have a use case where I want to validate that the digest algorithms of the SignedData are among selected ones. Thus I need access to them during parsing.


andy-goryachev:
should probably make a copy perhaps?

bcgit:
Merged with minor changes. The Set now returns the algorithm identifiers. Have added an equivalent method to CMSSignedData. Thanks.

PR #115, opened 2015-06-23
Correct a bug when adding a passphrase to a PGPSecretKey that was storing private data unencrypted

dallotTazTag:
In that case, the S2KUsage was never updated from SecretKeyPacket.USAGE_NONE to SecretKeyPacket.USAGE_SHA1


bcgit:
I've fixed this - it goes to USAGE_CHECKSUM though as a SHA-1 digest calculator will not necessarily be available.

PR #116, opened 2015-06-23
Add possibility to generate secret keys in GNU Divert-to-card format

dallotTazTag:
Allows to generate keys in the format used by GPG to store secret keys that are embedded on a smartcard.


dghgit:
I've added this as this is still a thing. There doesn't seem to be any decent way to test it at the moment though so any feedback in that area would be appreciated.

PR #129, opened 2015-12-28
Clone key in copy constructor

akwizgran:
The copy constructor for Blake2bDigest clones the new instance's own key rather than the key of the instance being copied.


bcgit:
Thanks. Applied.

PR #130, opened 2015-12-30
Clone personalization string in copy constructor

akwizgran:


bcgit:
I think this has now been fixed properly. Changes should appear on github shortly.

PR #131, opened 2015-12-31
Ensure key block is zero-padded when resetting Blake2bDigest.

akwizgran:
When Blake2bDigest is used with a key, the reset() method reuses the input buffer for the key block without zeroing the buffer. If the buffer contains input longer than the key, the key block will not be correctly zero-padded.


bcgit:
Nice work. Thanks, now merged.

PR #132, opened 2016-01-05
BLAKE2s implementation

str4d:
Based on the existing `Blake2bDigest`, conforming to the same API (despite the issue of some constructors requiring a `digestLength` in bits, and others in bytes).

The tests were passing in JUnit4, but I have not checked that my migration of them to BC's test harness was successful.


bcgit:
Merged with minor changes.

PR #137, opened 2016-03-26
Moved get*Name methods to public PGPUtil.

slawekjaranowski:
When we want to display info about some PGP object
it will be useful to have name insted of id.


bcgit:
Merged with minor changes. Thanks.

PR #139, opened 2016-04-11
Add test artifacts generated in Gradle build to gitignore

bdhess:
When running a Gradle build, a number of test artifacts are created in the `pg` directory.  Adding to gitignore.


peterdettman:
Merged, thanks.

PR #140, opened 2016-04-11
Fix issue with provider intialization during Gradle build

bdhess:
In general, the test execution logic seems to assume that setup and teardown happen immediately before/after the test is executed, and before any other tests run.  This is a bit of a hack to work around the `BC` provider being removed by another suite before these tests execute.  This seems consistent with other workarounds previously applied (see, e.g. the changes to SMIMEToolkitTest.java in 6db88cc).


peterdettman:
Merged, thanks again.

PR #142, opened 2016-04-25
Offer multiple IV-Prevention modes for different use-cases

sa-ChristianNiessner:
Hi,

this pull requests introduces multiple "Split Application Data Record" modes to prevent the "known IV weakness" in TLS 1.0.

The modes are:
* The "current" 1/n-1 record splitting mode which is still the default
* The 0/n split mode which was the default before the a23b7145c24d31554dbfc4d706366f28c7445fb0 commit
* An 0/n split mode which only applies for the first application data packet sent. I also implemented this mode because i've seen that my endpoints handle it this way.

As already mentioned the default mode keeps the current behaviour, but the behaviour can be changed when needed by using the 'setSplitApplicationDataRecordMode()' method.

What is the motivation behind this pull request:

We are using BC for handling 802.1x authentication requests. Those authentication requests are message-based, and it seems that most endpoints expect a 1:1 inner message / outer (TLS) messeage (packet) matching and don't support fragmentation of inner messages over multiple TLS packets.
There is one little sentence in one (very old but still to be supported) RFC:
https://tools.ietf.org/html/draft-kamath-pppext-peapv0-00#section-1.1

> The [PEAP] specification requires that EAP packets be tunneled within a
TLS channel in their entirety.

This does not explicit mention this fact but from our test all tested client implementations (Linux (inc. Android), Microsoft, and Apple (OSX / iOS)) can't handle the authentication messages properly when the default 1/n-1 record splitting mode applies.
With the 0/n splitting mode authentication works without any issues. I assume that those 'Zero-Length Fragments' are handled by the corresponding SSL lib's internally and so no fragmentation needs to be done by the authenticator.

During our tests we found out that most clients (only some microsoft versions don't sent those zero-length fragments et all) only send one "Zero-Length Fragment" before the first real application data frame. This is why we also offer this mode here. It also seems to be the behaviour for OpenSSL.

We expect that this 'message <> packet' matching also might be required by other use cases so we ask to implement this pull request into the main stream.

Thanks & Bye,
     Chris

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/tls/TlsProtocol.java
+++ b/core/src/main/java/org/bouncycastle/crypto/tls/TlsProtocol.java
@@ -45,6 +45,13 @@ public abstract class TlsProtocol
     protected static final short CS_END = 16;
 
     /*
+     * Different modes to handle the known IV weakness
+     */
+    protected static final short SADR_MODE_1_Nsub1 = 0; // 1/n-1 record splitting
+    protected static final short SADR_MODE_0_N = 1; // 0/n record splitting
+    protected static final short SADR_MODE_0_N_FIRSTONLY = 2; // 0/n record splitting on first data fragment only
+
+    /*
      * Queues for data from some protocols.
      */
     private ByteQueue applicationDataQueue = new ByteQueue();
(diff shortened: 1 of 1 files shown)


peterdettman:
Merged, with changes to the naming scheme only. Thanks for the patch. FYI, my recollection is that the 1/N-1 default in BC (originally it was 0/N) came from reading reports on the IETF TLS list, but I could be completely wrong on that. Good to support different options here in any case. Thanks again.

PR #149, opened 2016-07-21
GOST3411-2012 implementation

andrey-vasilyev:
Implementation of GOST R 34.11-2012 hashing function which is the replacement for GOST R 34.11-94 national standard of Russian Federation.

The implementation is well tested and optimized for speed. Unit tests are implemented also.


bcgit:
I've been reviewing this one - there's something odd going on though, I think I understand how it works relative to RFC 6898, but can you tell me what's going with the hmac examples offered in RFC 7836? There seems to be some sort of inconsistency there. Thanks.

andrey-vasilyev:
Yeah, it took me a while to figure out this one. For some unknown (to me at least) reason [the reference implementation in C language](https://www.streebog.net/) outputs the resulting digest in reverse order compared to what is given in RFC 6986 examples. I can only guess this is because in RFC 6986 the example messages themselves are given in reverse order! So when you get your first digest in HMAC using my implementation you have to reverse the resulting digest first and only then append it to the result of (K XOR opad) and of course reverse the resulting digest to get the same value as in the examples of RFC 7836. 

So if you believe it would be better for everyone to produce the same reversed result as the reference implementation does, I can easily modify my code.

bcgit:
Okay, the reversal makes sense as the way algorithm is defined you'd need to know the entire before constructing a digest otherwise... it'd be pretty unusual for someone to do that. Doing the reverse on the hash fixes the HMAC output as well. Funny thing though, similar thing with SHA-3. I've pushed the lightweight classes with some minor changes, should be appearing on github shortly.

andrey-vasilyev:
Great! Thank you. Nice job removing ByteArrayOutputStream! Guess I can close this one now.

bcgit:
And thank you!

PR #162, opened 2016-11-10
Don’t expose RSA private key in toString method

andybons:


bcgit:
I think I understand the motivation behind this one, but I'm afraid a lot of people regard this as a feature.
Having said that, since 1.8 the default JVM behavior also moved to this, so we might be able to get away with it.

bcgit:
This is now done - we've opted to print a fingerprint and the public key details for a private key. The fingerprint is derived from the public key as well, but should make it easier to match a private key with it's corresponding public key when going through logfiles.

PR #163, opened 2016-11-10
Fix ECDH bug

andybons:
Ensure that the EC computation is on the curve used by the private key, by using explicitly only
the public point and the private key.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/agreement/ECDHBasicAgreement.java
+++ b/core/src/main/java/org/bouncycastle/crypto/agreement/ECDHBasicAgreement.java
@@ -6,6 +6,7 @@ import org.bouncycastle.crypto.BasicAgreement;
 import org.bouncycastle.crypto.CipherParameters;
 import org.bouncycastle.crypto.params.ECPrivateKeyParameters;
 import org.bouncycastle.crypto.params.ECPublicKeyParameters;
+import org.bouncycastle.math.ec.ECCurve;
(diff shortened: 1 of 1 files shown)


peterdettman:
I am inclined instead to this alternative, which should achieve the same things:

```
   public BigInteger calculateAgreement(
        CipherParameters pubKey)
    {
        ECPublicKeyParameters pub = (ECPublicKeyParameters)pubKey;

        ECDomainParameters canonicalParams = key.getParameters();
        if (!canonicalParams.equals(pub.getParameters()))
        {
            throw new IllegalStateException("ECDH public key has wrong domain parameters");
        }

        // Always perform calculations on the exact curve specified by our private key's parameters
        ECPoint pubPoint = canonicalParams.getCurve().decodePoint(pub.getQ().getEncoded(false));
        if (pubPoint.isInfinity()) 
        {
            throw new IllegalStateException("Infinity is not a valid public key for ECDH");
        }

        ECPoint P = pubPoint.multiply(key.getD()).normalize();
        if (P.isInfinity())
        {
            throw new IllegalStateException("Infinity is not a valid agreement value for ECDH");
        }

        return P.getAffineXCoord().toBigInteger();
    }

```

Please let me know if this meets your concerns here.

It is somewhat unfortunate that there is then double-validation happening (existing check via ECPublicKeyParameters constructor), but let's sort out correctness first.

bcgit:
After much thought, I've merged in Peter's suggestion.

PR #164, opened 2016-11-10
Check DSA parameters in initSign() and initVerify() in DSASigner.java

andybons:


peterdettman:
As an alternative, these checks could be made in the DSAParameters constructor instead; is there a strong preference for checking it during these inits (instead/also)?

bcgit:
I still need to think about this one. The current patch may cause havoc with people dealing with legacy systems, although the idea itself isn't a bad one.

bcgit:
Closing off - we're looking at adding a "minimum security" mode, once we work out how to do that, I think checks like this can be captured there without causing people surprise.

PR #165, opened 2016-11-10
Fix wraparound counter bug which leaks plaintext and auth key

andybons:
If the plaintext/ciphertext length is larger than 2^36 - 32 bytes, the counter is wrapped and it leaks the plaintext and authentication key.

This issue and fix was originally authored by Quan Nguyen

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/modes/GCMBlockCipher.java
+++ b/core/src/main/java/org/bouncycastle/crypto/modes/GCMBlockCipher.java
@@ -25,6 +25,10 @@ public class GCMBlockCipher
 {
     private static final int BLOCK_SIZE = 16;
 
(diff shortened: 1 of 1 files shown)


peterdettman:
I believe this limitation is already enforced in the current code, except by counting blocks. See the `GCMBlockCipher.blocksRemaining` field and its usage. Could you please re-examine on that basis, and let us know if you think that insufficient/incorrect for some reason.

bcgit:
I've a feeling this might be based on an earlier version of BC, although if there is an error with blocksRemaining we would like to hear about it.

andybons:
This is indeed based on a change made to 1.52. I’ve asked my colleagues with more domain expertise (the ones who actually wrote the patches) to help with these PRs.

bcgit:
Closed as appears to be already dealt with.

PR #178, opened 2017-03-31
ISAACEngine: remove extra isaac() call

Adam-:
Also return bytes from from the end first, like many other
implementations.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/engines/ISAACEngine.java
+++ b/core/src/main/java/org/bouncycastle/crypto/engines/ISAACEngine.java
@@ -16,7 +16,8 @@ public class ISAACEngine
 {
     // Constants
     private final int sizeL          = 8,
-                      stateArraySize = sizeL<<5; // 256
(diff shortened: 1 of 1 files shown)


bcgit:
The patches change the known vector tests. I have reconfirmed the values against the authors own Java implementation as well.

Adam-:
Are we looking at the same implementation? http://www.burtleburtle.net/bob/java/rand/Rand.java is referenced on http://www.burtleburtle.net/bob/rand/isaacafa.html as the Java port created by the author.

If you read it, you'll notice it invokes Isaac() in its Init() method and then sets count=SIZE, so the next call to val() will return rsl[SIZE-1].

Your implementation calls isaac here: https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/engines/ISAACEngine.java#L184 after setting index=0. On the next call to returnByte(), it recalls isaac() due to index being 0. This causes the first 256 values generated to be skipped over and not used.

PR #179, opened 2017-04-02
Fix class documentation

mtausig:
Class documentation consisted partly of references to "EnvelopedData" instead of "EncryptedData". Probably a copy/paste error.


bcgit:
Thanks. Merged.

PR #181, opened 2017-04-06
add Constructor

VivleSoren:
GoppaCode.createCanonicalCheckMatrix(field, gp); take to much time for a constructor.
I want to save and load the McElieceCCA2PrivateKeyParameters in a file.

Here are my numbers as an example:
Intel i5-4570
a generated key pair with: McElieceCCA2Parameters(6624)
GoppaCode.createCanonicalCheckMatrix(field, gp); takes 30 seconds
new PolynomialRingGF2m(field, gp); takes 1,3 seconds


bcgit:
added, should appear soon.

PR #190, opened 2017-05-24
GOST3410-2012 implementation

MikeSafonov:
Implementation of GOST R 34.10-2012 signature algorithm which is the replacement for GOST R 34.10-2001 national standard of Russian Federation. [RFC 7091](https://tools.ietf.org/html/rfc7091.html)
Also added algorithms identifiers and names for using in CMS signature. 
Unit tests are implemented also.


bcgit:
Thanks. Will appear in158b03 on https://www.bouncycastle.org/betas (around the 30th May)

MikeSafonov:
Thank you!

PR #191, opened 2017-06-01
GOST34.10 2012  bug fix

MikeSafonov:
I apologize, but it seems that I did not correctly interpret RFC 7091. Test examples in RFC 7091 are given in little endian so we must reverse input hash-message.

Now the implementation has been successfully tested with [GostCrypto ](http://gostcrypto.com/) library


bcgit:
Thanks! This is now available in https://www.bouncycastle.org/betas

bcgit:
Now on https://www.bouncycastle.org/betas 158b04 or later

PR #192, opened 2017-06-02
Prevent NullPointerException during class loading in AES.class

Oja95:
Helps to avoid NullPointerException being thrown by `AES.class.getClassLoader().loadClass(className);`.
If class loader is null, it must've been loaded with bootstrap class loader.


Oja95:
See http://www.bouncycastle.org/jira/browse/BJA-676

bcgit:
Thanks for the suggestion and the report. This is should now be fixed. See https://www.bouncycastle.org/betas for the patched jar.

PR #195, opened 2017-06-16
DSS-1415: Support legacy encoding of ts tokens preserving cert ordering.

netmackan:
Regarding the discussion in the thread https://bouncycastle.org/devmailarchive/msg15273.html , the attached patch adds a parameter that can be used to get the same behavior as before BC 1.50 for instance when it comes to the ordering of the certificate in the output.

### DIFF
--- a/pkix/src/main/java/org/bouncycastle/tsp/TimeStampResponseGenerator.java
+++ b/pkix/src/main/java/org/bouncycastle/tsp/TimeStampResponseGenerator.java
@@ -1,5 +1,6 @@
 package org.bouncycastle.tsp;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.util.Date;
@@ -8,6 +9,7 @@ import java.util.Iterator;
 import java.util.Set;
 
 import org.bouncycastle.asn1.ASN1EncodableVector;
+import org.bouncycastle.asn1.ASN1InputStream;
 import org.bouncycastle.asn1.ASN1Integer;
 import org.bouncycastle.asn1.ASN1ObjectIdentifier;
 import org.bouncycastle.asn1.DERBitString;
@@ -20,6 +22,7 @@ import org.bouncycastle.asn1.cmp.PKIStatusInfo;
 import org.bouncycastle.asn1.cms.ContentInfo;
 import org.bouncycastle.asn1.tsp.TimeStampResp;
 import org.bouncycastle.asn1.x509.Extensions;
+import org.bouncycastle.cms.CMSSignedData;
 
 /**
  * Generator for RFC 3161 Time Stamp Responses.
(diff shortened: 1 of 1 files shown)


bcgit:
I've managed to reproduce this. The patch suggested solves the immediate issue but does not actually fix the problem.

It turns out what is actually happening is that TimeStampResp is using a DERSequence, rather than a DLSequence and this is resulting in a DER encoding of the CMS SignedData making up the token. I've fixed TimeStampResp to use a DLSequence and also added code to produce a definite length encoding of the signed data as some people will have issues parsing indefinite length data embedded in definite length objects.

The change you are after should appear on github shortly in TimeStampResponseGenerator. It's slightly different to the old code as it is trying to avoid introducing indefinite length data. It will work with the old version of TimeStampResp as well (which is also now fixed).

netmackan:
Nice, that was a bit simpler.
But does that mean that the time-stamp response will not be DER encoded?
Looking in RFC#3161 for instance in section 3.4 if I understand it the response message should be DER encoded so I am a bit worried if that requirement will not be satisfied. What is your thought?

bcgit:
Okay, 3.4 is actually about using it with MIME. Technically this would mean the response is not properly DER encoded because the BER status of the SignedData is still being recognised (so the SET structures in it are not sorted). As far as the outside world is concerned the response will appear to be DER encoded as a definite-length sequence is the same as a DER sequence.

I think this is okay, even if used with MIME, although the standard specifies DER, I think in this case the only reason for DER is prevent indefinite-length encodings appearing, particularly in the outer layer. Thinking about it again, we should probably leave TimeStampResp as DER explicitly.

netmackan:
So in that case if you will be reverting to continue using DER explicitly then when might still need to have the legacy encoding option as suggested in this PR in order to not break legacy systems that are (incorrectly) relaying on the ordering of the certificates. Would that be possible to include?

bcgit:
Well, it's little complicated, the thing is that the re-encoding of the ContentInfo body hides the fact the object isn't DER encoded... So DER for the outer wrapper doesn't have quite the same effect as it might. I think I see what you're worried about though, leave it with me, I need to think if there is some "elegant" way out of this.

bcgit:
Okay, I think I now have a way out of this we can all live with.

       TimeStampResponse initResp = tsRespGen.generateGrantedResponse(request, ..., new Date());

Will now return the signedData in the response intact so that:

initResp.getEncoded(ASN1Encoding.DL); 

will return the signedData in definite-length format, so without the SET objects being sorted. This is what you need to use for people looking for the legacy option.

initResp.getEncoded() 

On the other hand will return the DER encoding and so be fully conformant with the stated encoding requirements for use with MIME.

netmackan:
My tests are passing and this looks like a more clean solution. Thanks again!
Cheers,

PR #212, opened 2017-08-24
Use correct digest identifier

cbxp:
Otherwise generated certificate signatures are invalid


bcgit:
Thanks for the report. This is now fixed and up on https://www.bouncycastle.org/betas

PR #216, opened 2017-08-28
Allow use GOST3410 private key encoded as ASN INTEGER or OCTET STRING

4garbage:
like it alowed in ECGOST3410. 
OpenSsl encode private key as INTEGER


bcgit:
This has been added. Thanks.

PR #217, opened 2017-08-30
Ignore JavaDoc doclint errors which kill release under Java 8

rtyley:
The Bouncy Castle project has dozens of javadoc errors which doclint tolerated under Java 7 but are now fatal under Java 8. When releasing to Maven Central (using `gradle uploadArchives`) the `javadoc` task gets run, and will fail when running under Java 8, blocking the release.

The solution here was taken from Stephen Colebourne's blog: http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

See also: http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#quiet

Several of the javadoc errors in Bouncy Castle are valid errors - for instance:

```
core/src/main/java/org/bouncycastle/pqc/crypto/mceliece/McElieceCipher.java:160:
error: exception not thrown: java.lang.Exception
     * @throws Exception if the cipher text is invalid.
```

You can quickly see these by running `gradle javadoc` under Java 8.

cc @cwgit

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,14 @@ buildscript {
    }
 }
 
+if (JavaVersion.current().isJava8Compatible()) {
+    allprojects {
+        tasks.withType(Javadoc) {
+            options.addStringOption('Xdoclint:none', '-quiet')
+        }
+    }
+}
+
 allprojects {
    apply plugin: 'java'
    apply plugin: 'idea'


bcgit:
Thanks. Merged, should show up shortly.

mouse07410:
@bcgit the fix was only partial - it addressed Gradle builds, but not the Ant builds. Could you please fix the Ant build in the same way?

bcgit:
Will do, there's a couple of other things I need to sort out with ant as well. Looking into it.

bcgit:
One other thing, what JDK are you using?

mouse07410:
> One other thing, what JDK are you using?

```
$ java -version
java version "1.8.0_141"
Java(TM) SE Runtime Environment (build 1.8.0_141-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.141-b15, mixed mode)
$ 
```

I think something like
```
diff --git a/ant/bc+-build.xml b/ant/bc+-build.xml
index 295ba7485..a5346a29e 100644
--- a/ant/bc+-build.xml
+++ b/ant/bc+-build.xml
@@ -217,6 +217,7 @@
             <sequential>
             <mkdir dir="@{docsDir}" />
             <javadoc maxmemory="384m" sourcepath="@{srcDir}"
+                     additionalparam="-Xdoclint:none"
                      destdir="@{docsDir}"
                      windowtitle="Bouncy Castle Library ${release.name} API Specification"
                      header="&lt;b&gt;Bouncy Castle Cryptography Library ${release.name}&lt;/b&gt;">
```
might do the job. But I don't know if it's sufficient, i.e., whether other .xml files in the `ant/` directory need patching.

PR #218, opened 2017-08-31
Fix an incorrect of SM3WithSM2 signature algorithm.

ekszz:
The SM3withSM2 algorithm compute the signature using the following steps:

Set M’=ZA || M, where ZA= SM3(ENTLA || IDA || a || b || Gx || Gy || Ax || Ay) and IDA has a default 16-bytes length value {0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38},
then the final signature is SM2(SM3(M')).

In BC sources, there are two problems:
Firstly, the default IDA is not set;
Secondly, it computes the SM3 of TBSCertificate in SM2Signer class and GMSignatureSpi class repeatedly.

In this commit, it fixed those problems, and added a new sm_ca test case in CertTest class which comes from CFCA EV SM2 ROOT CA, the largest CA in financial sector in China.
By the way, it also changed the certificate of sm_sign test case to CFCA intermediate certificate of CFCA EV SM2 ROOT CA.


bcgit:
Thanks a lot for the report. I think we messed this one up a bit. Changes should appear on github shortly, there is now also a new beta on https://www.bouncycastle.org/betas

One question, SM2 Key Exchange also makes use of the IDA parameter. Does it have the same default value as for signature generation, or is it something different?

ekszz:
In SM2 Key Exchange, the IDA parameter uses "USER ID" which is the ASCII of e-mail in the official document. In the BC sources and test cases, it seems correct.

But the results of doKeyExchangeTestF2m() in class SM2KeyExchangeTest are inconsistent with the examples of official documents.
And I looked at another open source project (based on openssl, https://github.com/guanzhi/GmSSL/blob/master/test/sm2test.c), it also shows that the results are inconsistent with the official documents. So I also suspect that there is a typographical error in the official document example, but I'm not very sure about that.

ekszz:
By the way, sources in package org.bouncycastle.jcajce.provider.asymmetric.ec also have same problems need to be fixed :D

bcgit:
Thanks. Fortunately the rewrite means the JDK 1.4 version is no longer required, I have removed it.

Thanks for checking the key-exchange. Yes, as far as I can tell that example in the official document is two halves of two different examples - the hash in the middle is incorrect, I checked this against a couple of other implementations of SM3 as well with the same result.

PR #219, opened 2017-09-01
Added ECGOST oids to public/private keys

MikeSafonov:
Added gost oids to public/private keys for ECGOST 2001/2012.  The presence of identifiers allows to read the keys generated using [openssl gost-engine](https://github.com/gost-engine/engine)  and vice versa


bcgit:
I think this one is already done, bit of a strange one, I ran into a different but similar problem myself. Have a look at the current beta and let me know if the problem is still there.

MikeSafonov:
seems like all works fine

PR #243, opened 2017-10-16
Fix off by one error in SCrypt.

isomarcte:
According to https://tools.ietf.org/html/rfc7914#page-7 the cost parameter, N in the code, must be strictly less than 2^(128 * r / 8). In our case this is 65536. The argument checks formerly contained an off by one error checking that N was less than or equal to 65536 rather than strictly less than 65536.


peterdettman:
Thanks for the report. A quick look suggests to me that our scrypt implementation was for an early draft, so I will look at what changes it needs to match the final RFC, including the N check.

isomarcte:
@peterdettman should I close this or do you want to merge it? AFAIK, there is no down side to merging this while you perform your more complete review of the other Scrypt parameters and implementation.

bcgit:
Thanks, now merged. SCRYPT has been checked against RFC 7914 as well.

Apologies for the delay (Peter got dragged off by TLS...)

PR #254, opened 2017-11-13
Remove junit from prov build.gradle

zoonooz:
I am using the SpongyCastle and found out that junit was added to the classpath same as here.

To fix the problem and for later if SpongyCastle pull the change, I think `compile 'junit:junit:4.12'` should be removed from prov module and use only the testCompile dependency which defined in root project's gradle.


bcgit:
Merged. Thanks.

PR #261, opened 2017-11-28
initial implementation of Token Binding Extension TLS

inthirakumaaran:
Hi,
This PR consists the implementation of **Token binding extension** to TLS layer.This implementation is done on the client side and Token binding extension is added to client Hello.

There are two new classes introduced
1. TokenBindingExtension - contains the basic parameters to negotiate
2. NegotiatedTokenBinding - captures the negotiated parameters

Validation of negotiated parameters is happening in NegotiatedTokenBindingClass and processTokenBindingExtension method in Tls protocol class.

The EKM is added after establishing the master secret in Tls client protocol class.

Negotiated values are saved in NegotiatedTokenBinding class and that class is added as an attribute to security parameters.


peterdettman:
Thanks for the PR. Sorry for the delay, I'll try to get to it shortly.

inthirakumaaran:
okay :)

tls/src/main/java/org/bouncycastle/tls/NegotiatedTokenBinding.java
```diff
@@ -0,0 +1,62 @@
+package org.bouncycastle.tls;
+
+public class NegotiatedTokenBinding {
+
+    String selectedKeyParameter;
```
kasunbg: Can you verify whether the default access modifier suits here?

tls/src/main/java/org/bouncycastle/tls/AbstractTlsClient.java
```diff
@@ -228,6 +228,10 @@ public Hashtable getClientExtensions()
             TlsExtensionsUtils.addSupportedGroupsExtension(clientExtensions, supportedGroups);
         }
 
+        //add TokenBinding Extension
+        TokenBindingExtension tokenBindingExtension =new TokenBindingExtension();
+        TlsExtensionsUtils.addTokenBindingExtension(clientExtensions,tokenBindingExtension);
```
kasunbg: Better format the code.

tls/src/main/java/org/bouncycastle/tls/NegotiatedTokenBinding.java
```diff
@@ -0,0 +1,62 @@
...
+            throw new TlsFatalAlert(AlertDescription.unsupported_extension);
+        }
+        this.setMajorProtocolVerison(serverdata[0]);
+        this.setMinorProtocolVerison(serverdata[1]);
+        if(serverdata[3]==0){
+            this.setSelectedKeyParameter("rsa2048_pcks15");
+        }else if (serverdata[3]==1){
+            this.setSelectedKeyParameter("rsa2048_pss");
+        }else if (serverdata[3]==2){
+            this.setSelectedKeyParameter("rsa2048_ecdsap256");
```
kasunbg: Please use constants for key parameters.

tls/src/main/java/org/bouncycastle/tls/TokenBindingExtension.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+
+public class TokenBindingExtension {
```
kasunbg: class comments will make it easier to understand the code

tls/src/main/java/org/bouncycastle/tls/TokenBindingExtension.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public static final Integer rsa2048_pcks15 = 0;
+    public static final Integer rsa2048_pss = 1;
+    public static final Integer rsa2048_ecdsap256 = 2;
```
kasunbg: Should these be in upper case since these are constants? Do contrast with the token binding spec as well.

inthirakumaaran: spec always denote them in the simple letters

Issue #266, opened 2017-12-13
Performance optimization of SHA3 digest


xipki:

xipki:
duplicate the #267

PR #267, opened 2017-12-13
Performance optimization of SHA3 digest

xipki:
This pull request fixes the issue #266. With the new implementation I got 33% better performance compared to the original one to digest 200MB data.


peterdettman:
Great, thanks. I've merged this and then inlined the KeccakPermutation method fully for another large improvement.

PR #268, opened 2017-12-14
Brainpool curve names

super-horst:
BC 1.55 changed brainpool curve names, without any mention in the release notes ...
Please revert brainpool curve names, so they match rfc5639 again -> lowercase 'p' to uppercase 'P'
https://tools.ietf.org/html/rfc5639#section-3


bcgit:
Sorry, the curve names in the table need to be lower case only.

peterdettman:
Fixed the names returned from getNames() to match the names in RFC 5639. Lookup by name remains case-insensitive. Thanks for the report.

PR #273, opened 2017-12-22
SM2Signer-based signature cannot be reused after Signature.sign() wit…

xipki:
…hout reinitialization.


bcgit:
Thanks! Merged with minor revisions.

PR #283, opened 2018-01-26
fixing two special character encoding issues when testing on other locales

dheldt:
Hi,

we have 2 minor issues when building bc-java on a German windows system. Both come from some strings in test cases, containing special characters  (ä and π), which we encode as \u????. Afterwards all tests work fine on our systems.

I hope there is no problem with this pull request.

thanks & regards,

daniel


sake:
Instead of using the codepoint notation, the gradle file could force the compiler to interpret the source code as being UTF-8. 

```
compileJava.options.encoding = "UTF-8"
compileTestJava.options.encoding = "UTF-8"
```

bcgit:
Thanks for the update. These are now merged.

PR #319, opened 2018-03-23
fix a bug in SM4Engine.decrypt(byte[] in, int inOff, int inLen)

ZZMarquis:
if the parameter "inOff" is not zero, it can not be decrypted successully


bcgit:
Thanks for the fix. Applied!

PR #323, opened 2018-03-28
Explicitly check for short buffers in RSA Ciphers.

### LINKED ISSUE #322: RSA CipherSpi should check for short buffers
Cipher.doFinal() and friends state that Ciphers should throw ShortBufferException when given an output buffer that's too small to hold the output.  Other BC ciphers do this (eg, https://github.com/bcgit/bc-java/blob/master/prov/src/main/java/org/bouncycastle/jcajce/provider/symmetric/util/BaseBlockCipher.java#L1109), but RSA ciphers instead just try to fill in the output buffer and throw ArrayIndexOutOfBoundsException when the buffer runs out.  This also happens after bOut has been modified, so if the caller retries the call with a properly-sized buffer they'll get incorrect output.

flooey:
Cipher.doFinal() says that ShortBufferException should be thrown when
an insufficiently-long output buffer is provided.  Other BC ciphers
check for a too-short output buffer before beginning the requested
operation, but the RSA cipher doesn't, it just begins the operation
and throws ArrayIndexOutOfBoundsException.  Switch to checking for and
throwing ShortBufferException.

Fixes #322.

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/CipherSpi.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/CipherSpi.java
@@ -18,6 +18,7 @@ import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
 import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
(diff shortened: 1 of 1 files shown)


bcgit:
Thanks for the patch. Added!

PR #336, opened 2018-05-02
build.gradle: Use AnimalSniffer to check for Android 8 compatibility

Flowdalic:


dghgit:
We're now at gradle 9, so I'm closing this as out of date. However it's still not a bad idea even now, 1.85 had an Android API issue which could have been avoided if we had been using something like this. We've built a custom task using animal sniffer and fixed our current settings on Android API Level 26. Contribution acknowledged as well. Thanks and sorry for taking for ever.

PR #343, opened 2018-05-15
Initial implementation of RFC4998

veebee:
Implements RFC4998 - https://tools.ietf.org/html/rfc4998.

Note: the verification process does not check whether the Evidence Record embeds the revocation info related to the TSA certificate(s), as this was not specifically defined in the RFC (unless I'm mistaken).


mprechtl:
Will this implementation be part of a future version of Bouncy Castle?

Best,
Mike

bcgit:
I've reviewed the ASN.1 classes and checked in versions of those (I think they're now correct, but without test data it's hard to tell). Please update the rest of the submission in light of the new ASN.1 classes. 

We also need some tests in general.

bcgit:
Some of the asn1 classes now moved to asn1.tsp... DataGroup and some others added to org.bouncycastle.tsp

dghgit:
Closing this as an initial implementation is now in org.bouncycastle.tsp.ers The high-level ATS Chain and Sequence classes still need to be added, but it may require a slightly different approach to what is given in here.

Issue #345, opened 2018-05-15
Implement TLS v1.3, including X25519 and Ed25519, especially within X509 certs.


catbref:
Now that TLS v1.3 has been approved by the IETF can it be implemented into BC please?

Note that section 9.1 "Mandatory-to-Implement Cipher Suites" states that "A TLS-compliant application [...] SHOULD support key exchange with X25519 [RFC7748]."

You have some X25519-related code already and I can submit a pull request for most of the Ed25519/X509 aspect however I am new to the BC codebase so it would need serious reviewing.

peterdettman:
Work on Ed25519 is under way and TLS 1.3 is beginning shortly.

lapo-luchini:
There is something regarding Ed25519 in open-keychain/bouncycastle#2.

catbref:
Yes, and I've just now submitted a pull-request with code based on the above but reworked for more seamless integration with bcgit:master.
I'm not claiming my pull-request is good-to-go but could help save a bit of time.
One issue is that there could be duplication of X25519/Ed25519 math code.

vikram919:
@catbref 
>One issue is that there could be duplication of X25519/Ed25519 math code.

?

peterdettman:
Closing as TLS 1.3 is in place now.

We do not yet enable it for the "TLS" SSLContext though; you currently need to use "TLSv1.3" explicitly, or else enable it on the individual sockets/engines, or use jdk.tls.client.protocols/jdk.tls.server.protocols system properties.

Relative to TLS 1.2, it is currently missing session resumption. Relative to the full TLS 1.3 spec it is missing several of the new features, e.g. early data, half-close. Specific issues can be raised for these, although they are not likely to be on our roadmap for this year.

PR #347, opened 2018-05-16
RFC7748/Ed25519 sample implementation

catbref:
This is mostly based on work done by Valodim of openkeychain and, in turn, str4d for the base Ed25519 code. 

I've changed the code to remove the need for a separate provider and tried to integrate with BC in a more seamless way (e.g. support for OIDs in DefaultSignatureAlgorithmFinder and other places).

Needs serious code review!
Also block-renamed classes from EDDSA to RFC7748 as there is a lot of crossover of X25519/Ed25519, for example in X509 certs and RFC7748 is the base for both.


vikram919:
Thanks for your contribution, as ed25519 is RFC8032 edward's curve implementation and RFC7748 defines X25519 function as montgomery form of curve25519, Both curves are birational. After quickly looking at your changes, it seems like the conversion between ed25519 to x25519 is not implemented ? or if I'm wrong?

catbref:
This is going to be duplicating what I said in issue #345, maybe you'll have to close that issue or this PR to help manage comments, but anyway...

No, there's no function to convert between Ed25519 and X25519 as it's beyond my skillset.
I took the work done previously by Valodim and str4d and helped integrate it better with BC to the point of manipulating X509 certificates, etc. But when it came to establishing a TLS session requiring X25519 exchange I realised it would take a whole new layer of work I might not be able to complete.

bcgit:
Thanks for the suggestions/patches. With the introduction of PGP support, I think we're finally feature complete with this. The latest is in https://www.bouncycastle.org/betas 165b14 or later

PR #349, opened 2018-05-16
Use two-byte checksum for unencrypted secret keys

### LINKED ISSUE #348: Bouncycastle adds 20 bytes SHA1 hash after plain secret key instead of two bytes checksum
As pointed out [here](https://github.com/kazu-yamamoto/pgpdump/issues/23#issuecomment-389495218) and discussed on the mailing list (link pending), bouncycastle seems to falsely export secret keys when exporting them without a password. In that case, BC 1.59 seems to append a SHA1 hash after the secret key, instead of just a 2 bytes checksum.

I suspect, that only key export without a password is affected.

Valodim:
Fixes #348 by unconditionally using a 2-byte checksum instead of SHA1 if encryption algorithm is NULL.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKey.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKey.java
@@ -112,12 +112,12 @@ public class PGPSecretKey
 
             byte[]    keyData = bOut.toByteArray();
 
-            pOut.write(checksum(checksumCalculator, keyData, keyData.length));
-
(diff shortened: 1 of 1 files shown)


bcgit:
Thanks, this patch is actually correct and has been added.

neuhalje:
Is this actually merged?

```sh
git tag --contains  81bca0133f19d5efac4857e0da8b772002545157

error: no such commit 81bca0133f19d5efac4857e0da8b772002545157
```

PR #351, opened 2018-05-18
Make PGPUtil.pipeFileContents use buffer and not leak file handle

ecki:
This avoids generating a new byte buffer if one was specified. It also makes sure the file handle is closed in case a exception while reading/writing happens.

It is generally safe to suppress exceptions from FileInputStream.close(), but let me know if you want the safeClose() removed. In case the buffer copy is a security pre-caution and it should remain, it should be documented in source.


bcgit:
Thanks for the patch. Applied.

PR #356, opened 2018-05-30
Bugfix and improvement of CMP messages

xipki:


bcgit:
Thanks for the patch. I've merged most of this, with a minor revision to preserve the original API. I've added a getCertReqIdValues() to PollReqContent which just returns an array of CertReqId values.

xipki:
Thanks for the quick fix.
BTW, please merge all my entries in http://bouncycastle.org/contributors.html. You can use "Lijun Liao <https://github.com/xipki>" as my identifier.

bcgit:
Are you also "lijun.liao at rub.de" and "jun.liao at gmail.com"?

xipki:
Yes. It is me.

PR #362, opened 2018-06-07
Change return type of PGPSecretKey.getUserIds() to Iterator<String>

vanitasvitae:
`PGPPublicKey.getUserIds()` returns an `Iterator<String>`. `PGPSecretKey.getUserIds()` internally calls `.getUserIds()` of its public key, it however has the return type `Iterator`, which overlays (I'm not sure what the correct term is :D) the Iterators real generic type.

My PR changes the return type of `PGPSecretKey.getUserIds()` to `Iterator<String>` to match that of the public key.


bcgit:
Thanks for the patch! Merged.

PR #363, opened 2018-06-12
Only clear security paramaters on fatal alert

brandonweeks:
bbc1dab changed behavior even when a fatal alert isn't received.


peterdettman:
Clearing the security parameters after the handshake (in all cases) is intended behaviour. If you need to use those parameters at the end of a handshake (typically to export channel bindings or key material), you should override the notifyHandshakeComplete callback. There are examples in the tests.

If you believe that is insufficient for your use-case, please give more details. BTW, I couldn't see the relevance of the commit you linked to.

peterdettman:
@brandonweeks Great. In case you are unaware, there are some methods on TlsContext that might do what you are trying to use the master secret for:

    exportChannelBinding, with constants in ChannelBinding class (RFC 5929)
    exportKeyingMaterial, with labels in ExporterLabel class (RFC 5705)

These cover most usages of the master secret, and IMO should be preferred to making a copy of the master secret and re-implementing the logic externally. If you are implementing some other standard not covered by those, we might be interested in providing the functionality in BC too.

PR #370, opened 2018-06-18
Fix log message

vanitasvitae:
This PR fixes a log message in `BcPublicKeyDataDecryptorFactory`, which falsely states that the data cannot be *encrypted*, while in fact it could not be *decrypted*.


bcgit:
Thanks for the patch. Merged!

PR #385, opened 2018-07-26
maybe StringBuilder is better than StringBuffer as local variable type

ZZMarquis:
If jdk is 1.5on, maybe StringBuilder is better than StringBuffer as local variable type


peterdettman:
We still maintain compatibility for JDKs back to 1.1 and J2ME, and prefer not to keep different versions of each file unless strictly necessary. We would consider changing individual usages of StringBuffer, but only if a significant benefit could be demonstrated.

ZZMarquis:
@peterdettman OK

PR #407, opened 2018-09-21
Fix ed25519 ignoring the public key offset

TimoRoth:
The same issue also exists in the c# version of this. Should I make a separate PR there?


peterdettman:
Thanks, I'll fix the C# version too.

peterdettman:
Merged, and C# fixed too, thanks for the report.

Issue #413, opened 2018-10-02
Diffie-Hellman key finalization zero-padding incompatibility


alrossi:
We are developers working on the dCache storage system (https://www.dcache.org), which supports multiple I/O protocols for data access.  The system is implemented in Java and we rely on the BouncyCastle library to establish secure connections to our server.

One of the I/O protocols is called _xrootd_ (http://xrootd.org). We implement the _xrootd_ server in Java and our users utilize the _xrootd_ client, called _xrdcp_, provided by http://xrootd.org, to access the data. The client is implemented in C++.

When we switched from BC 1.46 to 1.50, we started to encounter this error:

```
javax.crypto.BadPaddingException: pad block corrupted
        at org.bouncycastle.jcajce.provider.symmetric.util.BaseBlockCipher.engineDoFinal(Unknown Source) ~[bcprov-jdk15on-1.50.jar:1.50.0]
        at javax.crypto.Cipher.doFinal(Cipher.java:2165) ~[na:1.8.0_141]
```

in about 0.5% of the transfers.

We traced the error to the implementation of this function, which changed between 1.46 to 1.5:

`org.bouncycastle.jcajce.provider.asymmetric.dh.KeyAgreementSpi.bigIntToBytes`

where leading-zero padding of the byte array was introduced. This padding was not present in the 1.46 version of this function.

We also discovered that the _xrdcp_ client uses the _openssl_ function

`DH_compute_key`

instead of

`DH_compute_key_padded`

and that is why it fails when used against our system with BC 1.50.

We verified that :

1. We can successfully use the existing _xrdcp_ client against our system with a custom BC jar which reverts to the old (non-padded) implementation of `bigIntToBytes`.
2. Fixing the _xrdcp_ client by replacing `DH_compute_key` with `DH_compute_key_padded` works against our system with the 1.50 BC jar provided by you.

Unfortunately, (2) is not really a solution, as replacing `DH_compute_key`  with `DH_compute_key_padded` makes the client incompatible with the existing server implementations.  Although if there are very strong arguments for using the padded version, we can present these arguments when speaking with the _xrootd_ developers.

For us, it would have been ideal if BC (like _openssl_) could support both the padded and unpadded key finalization/exchange. 

What do you think?

Thanks for your input.

dghgit:
Closing as resolved in BC 1.64.

PR #414, opened 2018-10-02
Add KeyAgreementNoPaddingSpi that implements old, unpadded bigIntToBytes

DmitryLitvintsev:
to support clients that use unpadded DH_comute_key (openssl). Add
"DHNP" KeyAgreement algorithm to invoke this code path

### DIFF
--- a/build15+
+++ b/build15+
@@ -6,7 +6,7 @@
 # otherwise it just creates the docs and class files.
 #
 
-if [ "${JDKPATH}" = "" ] 
+if [ "${JDKPATH}" = "" ]
 then
 	JDKPATH=/usr/lib/jvm/java-8-openjdk-amd64
 fi
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/DHNP.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/DHNP.java
@@ -1,0 +1,23 @@
+package org.bouncycastle.jcajce.provider.asymmetric;
+
+import org.bouncycastle.jcajce.provider.config.ConfigurableProvider;
+import org.bouncycastle.jcajce.provider.util.AsymmetricAlgorithmProvider;
+
+public class DHNP
+{
+    private static final String PREFIX = "org.bouncycastle.jcajce.provider.asymmetric" + ".dh.";
+
+    public static class Mappings
+        extends AsymmetricAlgorithmProvider
+    {
+        public Mappings()
+        {
+        }
+
+        public void configure(ConfigurableProvider provider)
+        {
+            provider.addAlgorithm("KeyAgreement.DHNP", PREFIX + "KeyAgreementNoPaddingSpi");
+            provider.addAlgorithm("Alg.Alias.KeyAgreement.DIFFIEHELLMANNP", "DHNP");
+        }
+    }
+}
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/dh/KeyAgreementNoPaddingSpi.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/dh/KeyAgreementNoPaddingSpi.java
@@ -1,0 +1,210 @@
+package org.bouncycastle.jcajce.provider.asymmetric.dh;
+
+import java.math.BigInteger;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.Key;
+import java.security.SecureRandom;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.Hashtable;
+
+import javax.crypto.SecretKey;
+import javax.crypto.ShortBufferException;
+import javax.crypto.interfaces.DHPrivateKey;
+import javax.crypto.interfaces.DHPublicKey;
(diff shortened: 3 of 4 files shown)


bcgit:
Try generateSecret("TlsPremasterSecret") - I think it will do what you want correctly now, rather than having to add another service.

DmitryLitvintsev:
Hi, 
Thank you for suggestion! I think I will need some guidance. My current server side code is 
```KeyAgreement _keyAgreement = KeyAgreement.getInstance("DH", "BC");
    ....
   byte [] iv = new byte[16];
   Arrays.fill(iv, (byte)0);
   final byte[] sharedSecret = _keyAgreement.generateSecret();
   SecretKey sessionKey = new SecretKeySpec(sharedSecret,
                                                0,
                                                16,
                                                "AES");
   IvParameterSpec paramSpec = new IvParameterSpec(iv);
   Cipher cipher = Cipher.getInstance(""AES/CBC/PKCS5Padding","BC");
   cipher.init(Cipher.DECRYPT_MODE, sessionKey, paramSpec);
   byte[] decrypted = cipher.doFinal(encrypted); // encrypted is byte array off the wire
```

`generateSecret.generateSecret("TlsPremasterSecret")` returns `SecretKey` , not a byte array. Passing `_keyAgreement.generateSecret("TlsPremasterSecret").getEncoded()` to `SecretKeySpec` does not help.

bcgit:
You might need to include an Arrays.copyOfRange(_keyAgreement.generateSecret("TlsPremasterSecret").getEncoded(), 0, 16) - the return value is the full agreement value, converted to a byte array, or is there some other issue?

DmitryLitvintsev:
Ah, the issue is that if I do this
```
SecretKey sessionKey = new SecretKeySpec(_keyAgreement.
                                                                          generateSecret("TlsPremasterSecret").
                                                                          getEncoded(),
                                                                          0,
                                                                          16);
....
```
I get the same error `pad block corrupted` in 0.5% of tries, that I get by using original code that a quoted in my previous comment.  (and also there is an issue https://github.com/bcgit/bc-java/issues/413). Using suggested `Arrays.copyOfRange` does not help too.

DmitryLitvintsev:
OK, I figured out what the issue might be here. We are discussing code on the master, whereas my server runs against BC 1.50

As far as I could see there is no handling of " TlsPremasterSecret" in BC 1.50. 
(tag 1rv50). Interesting that it just did not fail w/ NoSuchAlgorithmException.

DmitryLitvintsev:
Indeed. So, I added handling of "TlsPremasterSecret" in `KeyAgreementSpi.java` on  r1rv50
(I had to add `trimZeroes` function to `KeyAgreementSpi` to handle "TlsPremasterSecret" alrgorithm)
and it seems to *work fine* with the workaround you have suggested (use `generateSecret("TlsPremasterSecret")`). 

I have not tried master as we can't easily adapt to it, but all indication is that it will work. 

Residual question: it is possible to add this fix to existing supported versions? 

Thank you, 
Dmitry

bcgit:
That's a bit unexpected, what key agreement algorithm are you using?

DmitryLitvintsev:
Well, on  r1rv50 : 

```
[litvinse@orbis bc-java]$ git checkout r1rv50
Previous HEAD position was ef676b6f7... update
HEAD is now at f6f530b18... updates
[litvinse@orbis bc-java]$ find prov/ -name '*.java' -exec egrep -H 'TlsPremasterSecret' {} \; 
[litvinse@orbis bc-java]$  
```
(Nothing about TlsPremasterSecret). I am using:

```
_keyAgreement = KeyAgreement.getInstance("DH", "BC");
```

bcgit:
Try 1.59

DmitryLitvintsev:
Thank you. I believe I can close this pull request. I have found a working solution:
 a patched `KeyAgreementSpi` in 1.50 + `generateSecret("TlsPremasterSecret")`. We will make our custom BC jar and will be using it until we can upgrade to newer version.

DmitryLitvintsev:
I am re-opening this pull request only to solicit more guidance/help.. 

I have explored switching our application to BC 1.60. 

It fails when trying to handshake w/ another application (xrootd server written in C++). 

```
org.bouncycastle.jcajce.provider.asymmetric.util.ExtendedInvalidKeySpecException: unsafe p value so small specific l required
       at org.bouncycastle.jcajce.provider.asymmetric.dh.KeyFactorySpi.engineGeneratePublic(Unknown Source)
       at java.security.KeyFactory.generatePublic(KeyFactory.java:328)
       at org.dcache.xrootd.plugins.authn.gsi.DHSession.finaliseKeyAgreement(DHSession.java:177)
       ...
```

It is true that xrootd application sends us a short P. It is possible to by-pass this somehow? (we can't easily change the other application). 

I see that a proper constructor exists:

```
public DHParameters(
        BigInteger  p,
        BigInteger  g,
        BigInteger  q,
        int         m,
        int         l)
    {
        this(p, g, q, m, l, null, null);
    }
```

But I don't see a way of invoking it. 
(I may be completely on the wrong track here so any advice is greatly appreciated)

bcgit:
For future reference it's better to open a new issue.This got missed.

I've add "org.bouncycastle.asn1.allow_p_value" if set to true the test is not carried out.

Issue #415, opened 2018-10-03
Sphincs fails with ClassCastException when signature.initSign gets a PRNG as an input


kchalkias:
Although SPHINCS is a deterministic signing algorithm, if we invoke 
`signature.initSign(privateKey, SecureRandomImpl())` for some reason (i.e., in generic crypto libs that they don't know the algorithm a priori), then `sphincs\SignatureSpi.engineInitSign` will 
return a `ParametersWithRandom` object and not a `SPHINCSPrivateKeyParameters`.
Then `SPHINCS256Signer.init` function will try to cast this object to `SPHINCSPrivateKeyParameters`, which will obviously fail. 

In `sphincs\SignatureSpi`
```
protected void engineInitSign(PrivateKey privateKey) throws InvalidKeyException {
        if (privateKey instanceof BCSphincs256PrivateKey)
        {
            CipherParameters param = ((BCSphincs256PrivateKey)privateKey).getKeyParams();
            // HERE IS THE PROBLEM, SPHINCS is deterministic by its nature, 
            // Maybe just omit the PRNG?
            if (random != null)
            {
                param = new ParametersWithRandom(param, random); 
            } 
            digest.reset();
            signer.init(true, param);
```
Then, in `SPHINCS256Signer`
```
    public void init(boolean var1, CipherParameters var2) {
        if (var1) {
            this.keyData = ((SPHINCSPrivateKeyParameters)var2).getKeyData();
        } else {
            this.keyData = ((SPHINCSPublicKeyParameters)var2).getKeyData();
        }

    }
```
related stack trace:
```
java.lang.ClassCastException: org.bouncycastle.crypto.params.ParametersWithRandom cannot be cast to org.bouncycastle.pqc.crypto.sphincs.SPHINCSPrivateKeyParameters
	at org.bouncycastle.pqc.crypto.sphincs.SPHINCS256Signer.init(Unknown Source)
	at org.bouncycastle.pqc.jcajce.provider.sphincs.SignatureSpi.engineInitSign(Unknown Source)
	at org.bouncycastle.pqc.jcajce.provider.sphincs.SignatureSpi.engineInitSign(Unknown Source)
	at java.security.Signature$Delegate.engineInitSign(Signature.java:1186)
	at java.security.Signature.initSign(Signature.java:553)
```

bcgit:
Thanks for the report. This is now fixed and will appear in 161b12 later next week.

PR #416, opened 2018-10-03
SPHINCS-256 signatures are deterministic and no RNG is required.

kchalkias:
A solution to the https://github.com/bcgit/bc-java/issues/415 where Sphincs fails with ClassCastException when signature.initSign gets a PRNG as an input.
We could possibly delete code from `sphincs/SignatureSpi.java`, but just in case a newer version of Sphincs requires some RNG, I've changed the `SPHINCS256Signer` logic only. 

Note: In the meantime `SignatureSpi.java` has been updated in master, but we should also do in `SPHINCS256Signer` to handle direct access/invocation.


bcgit:
I can see the merit in this. Merged. Thanks.

PR #423, opened 2018-10-29
Documentation of Test S/MIME Sign Encrypt and Decrease in Memory Use for SMIME Decryption

sarah-mdv:
Documentation to be able to understand better how the decryption and signing work in the example test.


bcgit:
Changes incorporated, thanks.

Issue #457, opened 2019-02-04
CertificateFactory does not conforms to specifications


ziccardi:
Java documentation says:

> If the data in the input stream does not contain an inherent end-of-certificate marker (other than EOF) and there is trailing data after the certificate is parsed, a CertificateException is thrown.

However the way the `CertificateFactory` provided by BC work is:

1. if The stream does not starts with 0x30, a PEM input is supposed
2. If the PEM input is invalid, `null` is returned

Below two snippets showing the issue
```java
byte[] v = new byte[]{ 0x31, 0x11, 0x12,0x13,0x14 };
CertificateFactory cf = CertificateFactory.getInstance("X509", new BouncyCastleProvider());
Certificate c = cf.generateCertificate(new ByteArrayInputStream(v));
// instead of rising an exception, c is null
```

The same is for the snippet below:
```java
CertificateFactory cf = CertificateFactory.getInstance("X509", new BouncyCastleProvider());
Certificate c = cf.generateCertificate(new ByteArrayInputStream("Hello world\n".getBytes()));
// instead of rising an exception, c is null
```

In both the cases, the file is treated as `PEM` encoded (even if the first is clearly a binary) just because the first byte is not 0x30.

dghgit:
It turned there was still one unmet case on this after the PR was applied. This is now fixed as well. Up on https://www.bouncycastle.org/betas

PR #459, opened 2019-02-04
Changed PEMUtil so that a check is performed that the header/footer boundaries are present

ziccardi:
# Motivation
ISSUE: #457 

# Description
The `PEMUtil` class has been changed so that checks about the presence and validity of PEM boundaries are performed

# Tests
- [x] Run the test suite and verify that tests still pass
- [x] Test with a valid file
- [x] Test with the output of `openssl ... -text` command
- [x] Test with invalid binary file
- [x] Test with an invalid text file

The `mail:test` fails, but it was failing before I made this patch too. I didn't check what was wrong about it.

# Step to verify

Try the following code:
```java
byte[] v = new byte[]{ 0x31, 0x11, 0x12,0x13,0x14, '\n' };
CertificateFactory cf = CertificateFactory.getInstance("X509", new BouncyCastleProvider());
Certificate c = cf.generateCertificate(new ByteArrayInputStream(v));
System.out.println(c);
c = cf.generateCertificate(new ByteArrayInputStream("Hello world".getBytes()));
System.out.println(c);
```

Without the patch, it will print
```
null
null
```
With the patch an exception will be thrown:
```java
Exception in thread "main" org.bouncycastle.jcajce.provider.asymmetric.x509.CertificateFactory$ExCertificateException: parsing issue: Malformed PEM data: no header found
	at org.bouncycastle.jcajce.provider.asymmetric.x509.CertificateFactory.engineGenerateCertificate(CertificateFactory.java:240)
	at java.security.cert.CertificateFactory.generateCertificate(CertificateFactory.java:339)
	at test.main(test.java:94)
Caused by: java.io.IOException: Malformed PEM data: no header found
	at org.bouncycastle.jcajce.provider.asymmetric.x509.PEMUtil.readPEMObject(PEMUtil.java:124)
	at org.bouncycastle.jcajce.provider.asymmetric.x509.CertificateFactory.readPEMCertificate(CertificateFactory.java:67)
	at org.bouncycastle.jcajce.provider.asymmetric.x509.CertificateFactory.engineGenerateCertificate(CertificateFactory.java:231)
	... 2 more
```

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/x509/PEMUtil.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/x509/PEMUtil.java
@@ -8,43 +8,63 @@ import org.bouncycastle.util.encoders.Base64;
 
 class PEMUtil
 {
-    private final String _header1;
-    private final String _header2;
-    private final String _header3;
-    private final String _footer1;
-    private final String _footer2;
-    private final String _footer3;
-
-    PEMUtil(
-        String type)
+    /**
+     * Boundary class. Keeps track of the required header/footer pair for the
+     * current PEM object.
+     *
+     */
+    private class Boundaries
     {
-        _header1 = "-----BEGIN " + type + "-----";
-        _header2 = "-----BEGIN X509 " + type + "-----";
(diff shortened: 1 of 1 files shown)


bcgit:
Merged. Thanks!

PR #463, opened 2019-02-17
Add Ethereum flavor of IES engine

atoulme:
Ethereum DevP2P uses a slightly different approach to encoding IES by enforcing a common mac.

This class is used as part of the DevP2P RLPx library of Cava, but could potentially see some use in other projects.


bcgit:
Merged. Thanks!

PR #466, opened 2019-02-22
ignore the index

ZZMarquis:
ignore the index


peterdettman:
Merged, thanks for the fix.

PR #468, opened 2019-02-26
Added constructor to JceKeyTransRecipientInfoGenerator

sarah-mdv:
The constructor allows the user to directly provide the JceAsymmetricKeyWrapper directly as a parameter.


bcgit:
added with minor revisions. Thanks.

PR #472, opened 2019-03-05
 Add test for JournalingSecureRandom encryption 

sarah-mdv:
Test for coherence of output when recreating an encryption session with a journaling secure random


bcgit:
Merged with minor changes. Please note any contributed code must be under the BC license. I've assumed the license comment was an accident, if that was not the case please let me know and I will delete the file. Thanks.

PR #475, opened 2019-03-08
update comment

gabrielsroka:
the value on line 40 seems to be the product of 131 primes, not 132 primes.

also, 751 is not included, so it's really more like [3, 751)


peterdettman:
Merged with followup changes, thanks.

PR #486, opened 2019-03-29
Enable "NONEwithRSA" signatures with RSADigestSigner mechanism

oh2mqk:
This is same patch as the attachment on bcgit#485


oh2mqk:
Now RSADigestSigner works alike ...rsa.DigestSignatureSpi.
Specifically the derEncode() method works alike, and RSADigestSigner's constructor allows Null Digester to be used.

A side-effect is that any unknown digester will also produce unwrapped RSA signatures.
Should that be somehow detected?

bcgit:
Merged. Thanks. I've added a check to ensure the "null" hash can be parsed as a DigestInfo at least.

PR #488, opened 2019-03-31
Fix typo in javadoc comment

yoohaemin:
`AbstractTlsServer` seems like a base class for a TLS server.


peterdettman:
Merged, thanks.

PR #496, opened 2019-04-11
Fixed RFC 5280 / RFC 3280 NameConstraint checking for RDNs

kaoh:
The current implementation is comparing if the complete RDNs are matching between the permitted / excluded subtree and the certificate. This is not correct. The subtree must take into account which type of a RDN which should be matched. It is legal to add not restricted RDN to the subject or alternative names.

See:  https://tools.ietf.org/html/rfc5280#section-7.1
The test specification from https://csrc.nist.gov/CSRC/media/Projects/PKI-Testing/documents/PKITS.pdf give a valid example in 6.1.5.382 Valid DN nameConstraints Test1 EE:

```
permittedSubtrees:
directoryName: "ou=permittedSubtree1, o=Test Certificates 2011,
c=US"
```

It is OK to have a CN not restricted by the permittedSubtrees:

```
subject: "cn=Valid DN nameConstraints EE Certificate Test1, ou=permittedSubtree1, o=Test
Certificates 2011, c=US"
```

I have used the new DNS comparison algorithm from https://tools.ietf.org/html/rfc5280#section-7.1

The reason why this was never an issue so far might be that this extension is rarely used. 
But a new specification using NameContraints is SGP.22 https://www.gsma.com/newsroom/all-documents/sgp-22-v2-2-technical-specification/ 
See there section "4.5.2.1.0.3 EUM" where a name constraint is defined.

I have modified to allow a substring comparison for the serialNumber RDN. It makes sense to consider here substring matching similar to the IP or email handling.
Serial numbers are defining a range where e.g. a manufacturer is defining a common lot number where individual device can append a unique serial number.

So far BC is the only open Java crypto library I'm aware of supporting this specification because of the used elliptic curve modes (Brainpool)

I have modified the test case to match the fixed behavior.

I addition I have simplified the tooling: The mail, activation and junit libraries are included now. The license allows this. 
As long as no modern dependency fetching build tool is used this seems to be the simplest approach. The previous set-up was OS file system dependent and required 
tweaking of the build property file which cannot be committed in a changed state


bcgit:
Thanks. I'd like you to check what's been done - I've merged the differences in the PKIXNameConstraintValidator from the jce.provider package into the asn1.x509 one and changed the jce.provider one to simply invoke the other. A task that's been well overdue. Pretty sure everything now captured. 

I've kept the check in of the supporting libraries. We'll see how it goes.

bcgit:
Now up on https://www,bouncycastle.org/betas

kaoh:
Sorry, this took some days, because I had to add the jar manually to my Maven test project. The current beta does not work for me. The breaking change is in:

https://github.com/bcgit/bc-java/blob/aef80ff07ff9ef8ad6eab7e4ae4172aa6e9d5d67/core/src/main/java/org/bouncycastle/asn1/x509/PKIXNameConstraintValidator.java#L90

Now all unknown `GeneralName` types are throwing an exception, in the last version any unknown `GeneralName` was ignored:

I have here a 2.999.10 GeneralName (`registeredID` type) set in the `subjectAltName`extension. 
The RFC 5280 states:

>  Restrictions are defined in terms of permitted or excluded name
>    subtrees.  Any name matching a restriction in the excludedSubtrees
>    field is invalid regardless of information appearing in the
>    permittedSubtrees.  Conforming CAs MUST mark this extension as
>    critical and SHOULD NOT impose name constraints on the x400Address,
>    ediPartyName, or registeredID name forms.

So, the previous check was correct. Please remove the exception again.


Also the exclude has this issue: https://github.com/bcgit/bc-java/blob/aef80ff07ff9ef8ad6eab7e4ae4172aa6e9d5d67/core/src/main/java/org/bouncycastle/asn1/x509/PKIXNameConstraintValidator.java#L129

bcgit:
Okay, I think this is now fixed properly. The change should appear here shortly, I've uploaded a new beta as well. 162b15 or later. Let me know how it goes.

kaoh:
This works now.

bcgit:
Actually I think there's still a problem.

In IETF parlance 

"cn=Valid DN nameConstraints EE Certificate Test1, ou=permittedSubtree1, o=Test
Certificates 2011, c=US"

will encode in the order:

"c=US, o=Test Certificates 2011, ou=permittedSubtree1, cn=Valid DN nameConstraints EE Certificate Test1"

and the constraint for this would be:

"c=US, o=Test Certificates 2011, ou=permittedSubtree1"

The real problem is that the DN method was processing the subtree backwards, as it's working off the encoding it should be processing it forwards.

koh-osug:
You mean `if (!subtreeRdn.getTypesAndValues()[l].getType().equals(dnsRdn.getTypesAndValues()[l].getType()))` in  `org.bouncycastle.asn1.x509.PKIXNameConstraintValidator`? I added a note regarding the order, it this is now a problem a more involved algorithm must be used. Bit I think your example:

    "c=US, o=Test Certificates 2011, ou=permittedSubtree1, cn=Valid DN nameConstraints EE Certificate Test1"

and with constraint:

    "c=US, o=Test Certificates 2011, ou=permittedSubtree1"

should work here. The order matches here.

koh-osug:
Somewhere between the tested beta and the release this the `PKIXNameConstraintValidator` was not merged correctly. I have created a new pull request.

PR #498, opened 2019-04-18
it will terminate early if expected.length != supplied.length

ZZMarquis:
it will terminate early if expected.length != supplied.length


bcgit:
This will only terminate early if the two objects are identical.

ZZMarquis:
no, if expected and supplied are not identical and expected.length not equals the supplied.length, it will run this code:"return !Arrays.constantTimeAreEqual(expected, expected);", it will terminate early while call constantTimeAreEqual secondly.

bcgit:
I think I must have been having a slow day yesterday, I thought I might have been missing something. Yes, this will help improve the "messing with the attacker's head" value. Merged. Thanks.

PR #503, opened 2019-04-25
Implementation of the BLAKE2xs digest

### LINKED ISSUE #501: Xof implementation of Blake2s
Hi,

In the scope of my work, I need a Java implementation of an eXtendable-Output Function using Blake2s and after some research it doesn't seem to be available anywhere. As we're already using bouncycastle, I was wondering if you guys would be open for such an implementation knowing that the Blake2sDigest would need to be changed in order to change the internal tree configuration?

I have the time to work on this, I only want to make sure it won't be refused ;).

Thanks!

Gilthoniel:
This PR adds the implementation of the extendable output function of the BLAKE2s hash algorithm. The reference can be found here: https://blake2.net/blake2x.pdf.

Fixes #501


bcgit:
Merged with minor changes. Thanks!

PR #504, opened 2019-04-28
Return specific template type when possible

THausherr:
I suggest to change the return types to Store with the specific type template, this would remove the "unchecked method invocation" or "unchecked conversion" warnings. I've seen this done in other methods in recent version changes, although I can't remember specifics.


bcgit:
Thanks! Merged.

PR #506, opened 2019-04-29
Rejuvenate log levels

khatchad:
## Introduction

We are in the process of evaluating our [research prototype Eclipse plug-in](https://github.com/ponder-lab/Logging-Level-Evolution-Plugin) that "rejuvenates" log statement levels based on how "interesting" the enclosing methods are to the developers. The assumption is that methods that are worked on more and more recently by developers should have higher log levels (e.g., INFO as compared to FINEST). Our end goal is to reduce information overload, as well as alleviate developers from manually making log level changes.

The transformation decision is made by analyzing the "degree of interest" (DOI) values of enclosing methods for logging invocations. DOI value is a kind of real number for a program element which shows how developers are interested in it. It is computed from the interaction events between developer and element, such as developer edits the element. In this project, we compute the DOI using the project's git history.

We are looking for feedback on our tool from developers. If you can, we would appreciate if you can **comment on each of the transformations** in the case that this PR is not accepted. Of course, we would also love to contribute to your project.

## Settings

We have several analysis settings . We can vary these settings and rerun our if you desire. The settings we are using in this pull request are:

- Treat  CONFIG level as a category and not a traditional level, i.e., our tool ignores CONFIG log level (setting 1).
- Never lower the logging level of logging statements within catch blocks (setting 2).
- Never lower the logging level of logging statements within if statements (setting 3).
- Avoid log wrapping by disregarding logging statements contained in if statements mentioning log levels (setting 4).
- The greatest number of commits evaluated: 6000 (setting 5).

The head at time of analysis was: 4b133f18afe84364498f6f238a11e105e3e6df32

### DIFF
--- a/pkix/src/main/java/org/bouncycastle/pkix/jcajce/X509RevocationChecker.java
+++ b/pkix/src/main/java/org/bouncycastle/pkix/jcajce/X509RevocationChecker.java
@@ -396,7 +396,7 @@ public class X509RevocationChecker
 
         if (issuerList.isEmpty())
         {
-            LOG.log(Level.INFO, "configured with 0 pre-loaded CRLs");
+            LOG.log(Level.FINEST, "configured with 0 pre-loaded CRLs");
         }
         else
         {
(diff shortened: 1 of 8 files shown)


khatchad:
Sorry for the noise. Found a bug in this one.

saledouble:
This PR is outdated. We will submit a new one with a newer version of our tool.

khatchad:
Superseded by https://github.com/bcgit/bc-java/pull/623.

Issue #482, opened 2019-03-25
Update XMSS OIDs


securitykernel:
We previously contributed a draft-06 XMSS implementation to [Botan](https://botan.randombit.net) and we are currently [updating the implementation](https://github.com/randombit/botan/pull/1858) to conform to the final [RFC 8391](https://tools.ietf.org/html/rfc8391). We used BouncyCastle's implementation to generate fresh test vectors, as the RFC does not provide these. We noticed that BouncyCastle still uses draft-OIDs instead of the ones from the [final RFC 8391](https://tools.ietf.org/html/rfc8391#section-8) in [core/src/main/java/org/bouncycastle/pqc/crypto/xmss/DefaultXMSSOid.java](https://github.com/bcgit/bc-java/blob/738dfc0132323d66ad27e7ec366666ed3e0638ab/core/src/main/java/org/bouncycastle/pqc/crypto/xmss/DefaultXMSSOid.java). Updating these would increase compatibility of Botan's and BouncyCastle's implementation.

bcgit:
Now fixed.

PR #513, opened 2019-05-06
Update XMSS OIDs to final RFC 8391

securitykernel:
Implements #482.

OIDs for use with X.509 are not defined yet unfortunately, I could only find a [draft RFC](https://tools.ietf.org/html/draft-vangeest-x509-hash-sigs-03), so this PR does not touch them.


bcgit:
Thanks, merged with some modification. Looking through the RFC I also noticed an issue with the way public keys were being encoded (the OID is meant to be used as a prefix). I've also fixed that. Any feedback would be welcome.

securitykernel:
LGTM

bcgit:
Thanks. I'll mark this as done.

PR #521, opened 2019-05-18
Add chinese GM signature  alg support

Trisia:
- sha1 with sm2
- sha256 with sm2


peterdettman:
We're happy to add these, but could you please modify the patch to not include all the reformatting in GMSignatureSpi.

Trisia:
> We're happy to add these, but could you please modify the patch to not include all the reformatting in GMSignatureSpi.

OK I will commit again

PR #523, opened 2019-05-20
add signature alg support

Trisia:
- sha1 with sm2 
- sha256 with sm2

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/GM.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/GM.java
@@ -31,10 +31,10 @@ public class GM
             provider.addAlgorithm("Signature.SM3WITHSM2", PREFIX + "GMSignatureSpi$sm3WithSM2");
             provider.addAlgorithm("Alg.Alias.Signature." + GMObjectIdentifiers.sm2sign_with_sm3, "SM3WITHSM2");
             // SHA1 with SM2
-            provider.addAlgorithm("Signature.SHA1WithSM2", PREFIX + "GMSignatureFullSpi$sha1WithSM2");
+            provider.addAlgorithm("Signature.SHA1WithSM2", PREFIX + "GMSignatureSpi$sha1WithSM2");
             provider.addAlgorithm("Alg.Alias.Signature." + GMObjectIdentifiers.sm2sign_with_sha1, "SHA1WithSM2");
             // SHA256 with SM2
-            provider.addAlgorithm("Signature.SHA256WithSM2", PREFIX + "GMSignatureFullSpi$sha256WithSM2");
+            provider.addAlgorithm("Signature.SHA256WithSM2", PREFIX + "GMSignatureSpi$sha256WithSM2");
             provider.addAlgorithm("Alg.Alias.Signature." + GMObjectIdentifiers.sm2sign_with_sha256, "SHA256WithSM2");
 
             provider.addAlgorithm("Cipher.SM2", PREFIX + "GMCipherSpi$SM2");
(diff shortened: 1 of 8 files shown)


prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/GM.java
```diff
@@ -30,6 +30,12 @@ public void configure(ConfigurableProvider provider)
         {
             provider.addAlgorithm("Signature.SM3WITHSM2", PREFIX + "GMSignatureSpi$sm3WithSM2");
             provider.addAlgorithm("Alg.Alias.Signature." + GMObjectIdentifiers.sm2sign_with_sm3, "SM3WITHSM2");
+            // SHA1 with SM2
+            provider.addAlgorithm("Signature.SHA1WithSM2", PREFIX + "GMSignatureFullSpi$sha1WithSM2");
```
peterdettman: Why does this say GMSignatureFullSpi instead of GMSignatureApi? (same problem 3 lines below).

peterdettman:
BTW, you don't need to open a new issue when you make changes, just update your branch.

prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/GM.java
Trisia: this is my develop file , oh sory for that I`ll fix it

peterdettman:
I've merged this but have disabled the SHA1 code for the moment. We were only able to find test vectors for SHA256 and so far have been unable to confirm how SM2 should behave for hash algorithms that are not 256-bit output.

@Trisia If these changes were to support interoperability with other software, could you please advise what that is, and perhaps identify some test data for sm2sign_with_sha1 (or any of the larger hashes e.g. sm2sign_with_sha512) that we could use?

Trisia:
> I've merged this but have disabled the SHA1 code for the moment. We were only able to find test vectors for SHA256 and so far have been unable to confirm how SM2 should behave for hash algorithms that are not 256-bit output.
> 
> @Trisia If these changes were to support interoperability with other software, could you please advise what that is, and perhaps identify some test data for sm2sign_with_sha1 (or any of the larger hashes e.g. sm2sign_with_sha512) that we could use?

I needed to verify the signature in a PDF. The algorithm used was sha1 with SM2. Generally speaking, using SM2 would be used together with the SM3 hash algorithm. For compatibility I tried to implement it using the change algorithm. But I have not found the relevant GM standard to illustrate this algorithm, such as providing test cases and test data.

peterdettman:
We are having great difficulty in establishing any official support for digests apart from SM3. I think we're happy to support SHA256 since it seems obvious how that must work. For the others, we are still undecided, although SHA1 seems reasonably clear.

If the SHA1withSM2 PDF signature you are verifying is part of some business process, at this stage I would advise you to try and get that changed since it's likely to cause a lot of pain compared to just using e.g. SM3withSM2.

Of course we are still curious where the PDF came from and/or what software was used to sign it, since it may help us track down a standard or official definition.

Trisia:
> We are having great difficulty in establishing any official support for digests apart from SM3. I think we're happy to support SHA256 since it seems obvious how that must work. For the others, we are still undecided, although SHA1 seems reasonably clear.
> 
> If the SHA1withSM2 PDF signature you are verifying is part of some business process, at this stage I would advise you to try and get that changed since it's likely to cause a lot of pain compared to just using e.g. SM3withSM2.
> 
> Of course we are still curious where the PDF came from and/or what software was used to sign it, since it may help us track down a standard or official definition.

I addede ASN.1 DER Real type should I close this pull request and create new one?

PR #541, opened 2019-06-17
Memory optimisation of ARGON2 implementation

dbusche:
Hash computations allocates far to much temporary memory exhausting all system resources and requiring multiple GC runs per hash computation.

A memory profiler easily reports this issue:
unoptimized
An optimized version reduces the temporary memory consumption in this example by nearly a factor of 10:
optimized


bcgit:
Thanks for the patches. Merged with minor changes - will appear in 163b08 on https://www.bouncycastle.org/betas in a couple of days.

bcgit:
163b08 is now up.

PR #543, opened 2019-06-18
Fixed the InvalidKeyException message when initializing EDEC Signatur…

hitchan:
…eSpi from public to private when initializing for signing.


bcgit:
Thanks for the patch! Merged.

PR #549, opened 2019-06-26
added "cancel" functionality into DTLS process

fm-DmitryShyryayev:
Added "cancel" feature for DtlsClientProtocol and DtlsServerProtocol in tls layer

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/tls/DTLSClientProtocol.java
+++ b/core/src/main/java/org/bouncycastle/crypto/tls/DTLSClientProtocol.java
@@ -13,9 +13,19 @@ import org.bouncycastle.util.Arrays;
 public class DTLSClientProtocol
     extends DTLSProtocol
 {
+    private DTLSRecordLayer recordLayer;
+
     public DTLSClientProtocol(SecureRandom secureRandom)
     {
         super(secureRandom);
+    }
+
+    public void cancel() throws Exception
+    {
+        if (recordLayer != null)
+        {
+            recordLayer.close();
+        }
     }
 
     public DTLSTransport connect(TlsClient client, DatagramTransport transport)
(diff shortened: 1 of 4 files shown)


avenema:
Would love to see feedback on this. There doesn't seem to be another way that I can see to abort an active DTLS handshake.

peterdettman:
The first thing to note is that you are using the legacy TLS API (which I've recently explicitly deprecated). Please move to the new API (org.bouncycastle.tls) in the bctls jar, at least for purposes of testing and establishing how this cancellation should work.

The new API already supports a handshake timeout (override TlsPeer.getHandshakeTimeoutMillis()), in case that's all you're trying to do.

The patch supplied here doesn't really achieve much, because a thread blocked on a socket read won't be interrupted. Also, you can't add per-connection data to DTLSClientProtocol. The only real way to "cancel" reliably is to close the DTLSTransport (socket) from a separate thread. There was a bug with that in the legacy API, not sure if fixed there, but fixed in new API. We'll look at backporting that specific fix to legacy Java and C# APIs.

If you do still want a "cancellation" mechanism we could add an isCancelled method to TlsPeer.

avenema:
Is there any harm in merging this in? It's been in use for years now and has proven to be a very reliable way to abort the handshake externally. It's obviously preferable for us to continue to use a proven solution over re-writing and de-stabilizing our DTLS integration with BouncyCastle.

Even with the new API, the internal timeout isn't sufficient to address the use case, which requires cancelling the handshake based on some external trigger.

avenema:
To clarify the use case, the problem we ran into is that a handshake timeout (or user cancelled) would result in a blocked thread. Over time, the number of blocked threads increases, so we needed to add a mechanism by which threads stuck in the loop could be exited.

It sounds like the fix you mentioned regarding closing the DTLSTransport would solve the issue in a perhaps more graceful manner. At least in C#, closing the DTLSTransport from a separate thread doesn't seem to unblock the DTLSClient/ServerProtocol.

peterdettman:
I've pushed changes that are very similar to the proposed PR, except that the `cancel()` method is on `TlsPeer` (i.e. the user-provided `TlsClient`/`TlsServer` instance).

The changes were applied in both the legacy API (org.bouncycastle.crypto.tls) and the new one (org.bouncycastle.tls), but please be aware that we plan to completely remove the legacy API after the upcoming release (1.63).

avenema:
Thank you, @peterdettman!

Understood regarding the legacy API removal plans. Is it possible to have this change applied to the C# SDK, too? https://github.com/bcgit/bc-csharp/pull/207

peterdettman:
Yes, it will go into C# too.

avenema:
Thank you, @peterdettman!

peterdettman:
Fixed in 1.63.

PR #553, opened 2019-07-03
Give more precise message for corrupted stream error  in readLength()

sarah-mdv:
Give more information when errors are thrown.


bcgit:
Thanks. Actually we had the same thought, so there's already an updated error message. With the unexpected EOF the extra data won't help - you really need to tail the last bytes of the file to see the problem and usually it will be because someone has truncated the data somewhere else.

PR #564, opened 2019-07-22
Fixed name constraint comparison

koh-osug:
The `PKIXNameConstraintValidator` was most likely not merged correctly, it is now using again the old behavior where all subtrees and DNs must be identical. This is not according to https://tools.ietf.org/html/rfc5280#section-7.1

This is related to this pull request:

https://github.com/bcgit/bc-java/pull/496

### DIFF
--- a/core/src/main/java/org/bouncycastle/asn1/x509/PKIXNameConstraintValidator.java
+++ b/core/src/main/java/org/bouncycastle/asn1/x509/PKIXNameConstraintValidator.java
@@ -325,41 +325,35 @@ public class PKIXNameConstraintValidator
             return false;
         }
 
-        for (int j = 0; j < subtree.size(); j++)
+        for (int j = subtree.size() - 1; j >= 0; j--)
         {
             // both subtree and dns are a ASN.1 Name and the elements are a RDN
             RDN subtreeRdn = RDN.getInstance(subtree.getObjectAt(j));
-            RDN dnsRdn = RDN.getInstance(dns.getObjectAt(j));
-
-            // check if types and values of all naming attributes are matching, other types which are not restricted are allowed, see https://tools.ietf.org/html/rfc5280#section-7.1
-            if (subtreeRdn.size() == dnsRdn.size())
-            {
-                // Two relative distinguished names
-                //   RDN1 and RDN2 match if they have the same number of naming attributes
-                //   and for each naming attribute in RDN1 there is a matching naming attribute in RDN2.
-                //   NOTE: this is checking the attributes in the same order, which might be not necessary, if this is a problem also IETFUtils.rDNAreEqual mus tbe changed.
-                // use new RFC 5280 comparison, NOTE: this is now different from with RFC 3280, where only binary comparison is used
-                // obey RFC 5280 7.1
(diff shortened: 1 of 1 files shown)


koh-osug:
Any updates here?

bcgit:
I'm not so sure about this one, this was something that actually got fixed after the merge. Working backwards through the DN would be correct if it was an IETF string representation, but the code actually uses the encoding. Do you have an example this fails for? It would help if I had a clearer idea of what the problems, if any, there were. Thanks.

koh-osug:
Yes. I attach an example from the SGP.26 specification which is for the SGP.22 specification used as standard for all eSIM provision systems (https://www.gsma.com/newsroom/resources/sgp-22-v2-2-technical-specification/):

This is the signing certificate of the EUM (eUICC manufacturer = SIM manufacturer):
```
openssl x509 -in CERT_EUM_ECDSA_NIST.txt -text
Certificate:
    Data:
        Version: 3 (0x2)
        ...
        X509v3 extensions:
       ...
            X509v3 Name Constraints: critical
                Permitted:
                  DirName: O = RSP Test EUM, serialNumber = 89049032
   ....
```

Here the eUICC certificate signed by the EUM:

```
 openssl x509 -in CERT_EUICC_ECDSA_NIST.txt -text
Certificate:
    Data:
        Version: 3 (0x2)
       ...
        Subject: C=DE, O=RSP Test EUM/serialNumber=89049032123451234512345678901235, CN=Test eUICC
 ...
```

Here the name constraint of the EUM defines that any signed certificate must have a subject coming from the organisation `O = RSP Test EUM` and starting with a serial number of `serialNumber = 89049032` (support for serial numbers where already with a previous patch).

The eUICC certificate fulfills this requirement by coming from this organisation and is starting with the serial number. In addition the eUICC certificate defines other subject fields, here the country `C=DE` and a common name `CN=Test eUICC`. This is valid because the name constraint is only restricting the permitted subject names, not additional ones. The current BC implementation fails here. I quote here a different statement from RFC 5280  https://tools.ietf.org/html/rfc5280#section-4.2.1.10:

> Restrictions apply only when the
   specified name form is present.  If no name of the type is in the
   certificate, the certificate is acceptable.

The specified name forms are here the `O` and the `serialNumber`, not the `C` and `CN` why the name constraint check must not fail.

[CERT_CI_ECDSA_NIST.txt](https://github.com/bcgit/bc-java/files/3534182/CERT_CI_ECDSA_NIST.txt)

[CERT_EUM_ECDSA_NIST.txt](https://github.com/bcgit/bc-java/files/3534184/CERT_EUM_ECDSA_NIST.txt)

[CERT_EUICC_ECDSA_NIST.txt](https://github.com/bcgit/bc-java/files/3534188/CERT_EUICC_ECDSA_NIST.txt)

bcgit:
Okay, I think this is now "fixed" (by that I mean the above works and so do the NIST tests). Let me know how you go with it. The fixed version is up on https://www.bouncycastle.org/betas

koh-osug:
I have verified the beta and this works. Thanks.

bcgit:
Thanks.

PR #565, opened 2019-07-22
Issue 552 Performance problem when signing and with DNS resolve problems for localhost

kimmerin:
avoid long delays if there are DNS issues resolving the host's name while unnecessarily generating a message-id that is not needed here because the MimeMessage is only used to generate and copy MIME-headers for a given body part.


dghgit:
Thanks, now merged and up in https://www.bouncycastle.org/betas

Issue #528, opened 2019-06-04
CMSAuthEnvelopedData and CMSAuthEnvelopedGenerator java classes have only package access


wolftobias:
The two java files CMSAuthEnvelopedData and CMSAuthEnvelopedGenerator have only package access and it seems the implementation is not finished yet.
Background is I need to create a 1.2.840.113549.1.9.16.1.23 authEnvelopedData ASN1 structure and I tried to solve this with bouncycastle but the two classes are not really available for usage.

peterdettman:
Yes, these are placeholder classes for implementing RFC 5083 (and 5084). Unfortunately considerable work remains before they would be usable.

MadMud:
Hi,
I have written these for work. If there is interest I can ask if I am allowed to create a PR from this...
If not I can at least provide test vectors.

If memory serves you need to also add GCM in AES. The Algorithms are there but the CMS codec things are not there.

wolftobias:
If you could provide some input, I would be very appreciate.

bcgit:
@MadMud PR would be welcome providing you're happy with any code going out using the BC license. Test vectors are extremely important. Thanks.

wolftobias:
@MadMud Are there any new regarding your PR?

MadMud:
I have not heard back from my contacts at work. I think I will proceed with rewriting what is needed without my code from last year...

bcgit:
Ah, okay. This is starting to sound a bit grey. In this case, are you able to send some test vectors first? If we have those we can at least pull a parser together and probably a generator as well. At least then it would be clear that any subsequent contributions from you was on code that was external to your organization.

wolftobias:
@MadMud Now it`s your turn ;-), what is you plan? Are you going to support this?

MadMud:
Hi again, 
Had to care for a sick child for a week, but now I again have some free time :)

Anyway, I am unsure in which format to dump testvector on here. For starters lets got with [1], the key is [2] and the cleartext is "auth-enveloped data".

I propose to advance in one of two ways:
a) I can implement the basic functionality separate from the existing CMSEnvelopedDataStreamGenerator  et. al. This will involve quite a bit of code duplication, since most of the code from there needs to run also in the authenveloped case, We can discuss on how to remove this duplication in a second step.
[...]

PR #567, opened 2019-07-24
for #528 Basic AuthEnveloped Content Encryptor, inlcuding attributes. 

MadMud:
Hi, creating a PR for now. 
We might want to collect a bit more wore before we merge, but I wanted to easily point to my commit.

Basic AuthEnveloped Content Encryptor, inlcuding attributes. 
For now AES/GCM only, others can easily be added.


bcgit:
Just looking at this one now... there will be a couple of (smallish) changes...

MadMud:
Hi,
Sure thing! I have since changed the auth attributes handover to accommodate the authenveloped CMS generators. 
However these changes contain a bit of refactoring of the ASN.1 pojos and CMS parsers (authenveloped extends enveloped). 
I realize this is a bit of design change, so would you like me to put it in this PR or into a new one?
Cheers,
Matthias

bcgit:
I've synced up the latest. It should appear on github shortly.

bcgit:
Erm, everyone's noticed the work that was done on CMSAuthEnvelopedDataGenerator and CMSAuthEnvelopedData a couple of weeks ago haven't they?

MadMud:
On my mobile rightniw,but are referring to the merger of my pull request #528 ? That is in direct lint to what is in my fork now.
 At the time we wanted a way of producing authenveloped with authed attributes, so I did that first

MadMud:
Ok merge is done. Please have a look and comment on wether or not we want to proceed using Authenveloped extends enveloped

bcgit:
No, I wouldn't have AuthEnveloped extend Enveloped, ASN.1 is a lot of things, but OO it isn't. The two structures really different.

MadMud:
Hmm,
I generally agree, but in this case all of the properties in enveloped are in auth enveloped. Also in the same order. Have you looked at the code?
Anyway,the alternative is to copy all the code from enveloped an either all generators or extract an interface. 
Cheers,
Mad

dghgit:
Thanks for suggestions, marking this as closed as I think this has finally iterated to something workable. The most difficult thing to deal with properly was the manner in the mac gets treated. I think the current approach will work for streaming as well, but that's something we'll need to look at later.

PR #570, opened 2019-07-31
Add Zuc StreamCiphers and Macs

tonywasher:
Lightweight Implementation of Zuc-128 and Zuc-256 stream ciphers, plus associated Macs


bcgit:
Thanks for the PR. Merged with some minor changes. I've added JCE wrappers as well and uploaded it to https://www.bouncycastle.org/betas beta 163 b 12 or later.

PR #574, opened 2019-08-05
Don't try to encode TS token when null and DL encoding

mlundblad:
Avoid NPE when encoding response with no timestamp
token and definate length encoding is used.


bcgit:
You might have to tell me what you're trying to do here. The impression I get is you have a situation where you need to use DLSequence(). Is that correct?

bcgit:
Thanks for the patch, merged with minor changes.

PR #597, opened 2019-09-24
Added logic to identify private key type and get JCE private key from file

suraj0208:
The idea is to first read file as PEM first.

1. If we're able to read it as PEM, we get one of Bouncy Castle's objects wiz. PEMKeyPair, PrivateKeyInfo, PEMEncryptedKeyPair, PKCS8EncryptedPrivateKeyInfo. Depending on it we can Identify it as PKCS#1, PKCS#8, encrypted or not and get java.security.PrivateKey object. see getPrivateKeyType(Object bcPrivateKey) from the commit.
2. If we're not able to read it as PEM, fallback to reading it as DER and read key file in the form of ASN1 sequences. Depending on the number of sequences we can identify the type as PKCS#1, PKCS#8, encrypted or not and get java.security.PrivateKey object. see getPrivateKeyType(ASN1Sequence asn1Sequence) from commit.

### DIFF
--- a/pkix/src/main/java/org/bouncycastle/pkcs/JCEPrivateKeyFactory.java
+++ b/pkix/src/main/java/org/bouncycastle/pkcs/JCEPrivateKeyFactory.java
@@ -1,0 +1,483 @@
+package org.bouncycastle.pkcs;
+
+import org.bouncycastle.asn1.ASN1InputStream;
+import org.bouncycastle.asn1.ASN1Primitive;
+import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERNull;
+import org.bouncycastle.asn1.pkcs.PKCSObjectIdentifiers;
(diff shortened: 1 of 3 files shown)


bcgit:
I'll take a look. One thing this can't go in the org.bouncycastle.pkcs package as it is using JCA/JCE functionality. We need to think about somewhere else to put it.

suraj0208:
@bcgit 
Can you review it once again so that I can address all the review comments (if any) in only one or two commits? so far there's only one comment

suraj0208:
@bcgit a gentle reminder :)

dghgit:
Apologies for the delay, but the time has finally come... now merged with revisions as:

pkix/src/main/java/org/bouncycastle/openssl/jcajce/JcaPrivateKeyReader.java

Now up on https://www.bouncycastle.org/betas will ship in 1.85

PR #632, opened 2019-12-10
Handle null random values in PSSSignatureSpi

brandonweeks:
Uses the same pattern as SignatureSpi to handle null random parameters
passed into engineInitSign().

[JDK-8216039](https://bugs.openjdk.java.net/browse/JDK-8216039) introduced additional calls to engineInitSign() with null
random values which is causing test failures for us.


Acinyx:
This fixes https://github.com/bcgit/bc-java/issues/633

amitmun:
I've applied this fix locally on top of r1rv61 and built a custom JAR.
I can confirm this solved the issue.

Chr3is:
We're facing this issue too with BC 1.64 and JDK 11.0.5. We had to pin the java version to 11.0.3+7-1. Would be nice if this PR gets merged as soon as possible to avoid the jdk pin.

brandonweeks:
This change was merged in 640f6cba8362f7a28a09e457af86e4fd6724c190.

PR #638, opened 2019-12-31
Add the implementation of SipHash128, and extend the test of SipHash.

xipki:


bcgit:
Thanks for the patch. Merged!

Issue #646, opened 2020-01-18
Wrong class names registered in BC provider


a--v--k:
For some reasons I've written a unit test which checks if all classes, registered in BC provider by default, can be accessible and its instances can be created.

Key [AlgorithmParameterGenerator.1.2.804.2.1.1.1.1.1.3.5.1] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128 org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128.<init>()

Key [AlgorithmParameterGenerator.1.2.804.2.1.1.1.1.1.3.5.2] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen256 org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen256.<init>()

Key [AlgorithmParameterGenerator.1.2.804.2.1.1.1.1.1.3.5.3] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen512 org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen512.<init>()

Key [AlgorithmParameterGenerator.ARIACCM] contains wrong class name: org.bouncycastle.jcajce.provider.symmetric.ARIA$AlgParamGenCCM

Key [AlgorithmParameterGenerator.ARIAGCM] contains wrong class name: org.bouncycastle.jcajce.provider.symmetric.ARIA$AlgParamGenGCM

Key [AlgorithmParameterGenerator.DSTU7624] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128 org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128.<init>()

Key [AlgorithmParameterGenerator.GOST3410] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.asymmetric.gost.AlgorithmParameterGeneratorSpi 

Key [AlgorithmParameterGenerator.OID.1.2.804.2.1.1.1.1.1.3.5.1] contains class which failed on instance creation: org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128 org.bouncycastle.jcajce.provider.symmetric.DSTU7624$AlgParamGen128.<init>()

[...]

PR #647, opened 2020-01-18
#646 Wrong class names registered in BC provider

a--v--k:
Attempt to fix wrong class names registered in BC provider


bcgit:
Thanks for the patch. Merged with minor changes.

PR #653, opened 2020-01-30
Add support for explicit entropy source in NTRUEncryptionKeyGenerationParameters

hf:
It was not possible to use any other entropy source other than `CryptoServicesRegistrar#getSecureRandom()`. 

Even worse, the provided parameters use the `SecureRandom` from the registrar **present at class load time** which makes it difficult to customize the entropy source used by NTRUEncrypt.

Change is backward compatible, tho I'm planning on opening another PR that deprecates the existing parameters and uses another method of obtaining the standard NTRU parameters where one can specify the entropy source at generation time.


bcgit:
Thanks for the patch. Now merged with minor changes. Will show up in next beta (165b10).

Issue #655, opened 2020-02-09
serial number in test suite vulnerable to word shearing


whitingjr:
The assignment of a long is vulnerable to word shearing and results in code that is not correct. People are bound to copy the code from the test suite and this will be a source of a bug for users.

PR #656, opened 2020-02-09
Swapped out to use JRE implementation of AtomicLong.

whitingjr:
This PR looks to solve #655


bcgit:
Just in case I'm missing something here, what is your understanding of the "synchronized" keyword?

whitingjr:
The synchronized block does make the increment atomic. But the assignment/write means a reader *may* read only one updated word of the long. Long being vulnerable because the type is two word.
 It's why the replacement method getAndIncrement should be used. The increment and write is guaranteed to be atomic. if I'm not mistaken.

PR #658, opened 2020-02-12
Avoid using Android's default BC provider

### LINKED ISSUE #657: BouncyCastleProvider can't be used on Android without modifying the system provider list
I believe it should be possible to use a `java.security.Provider` instance standalone (without making changes to the system provider list) by passing an instance of the provider to factory methods such as [`Cipher.getInstance(String, Provider)`](https://developer.android.com/reference/javax/crypto/Cipher.html#getInstance(java.lang.String,%20java.security.Provider)).

On a desktop JVM this works fine - I can pass an instance of `BouncyCastleProvider` (without registering it as a system provider) and everything works as expected - but it doesn't on Android. This is because:

1. Android bundles an old version of BouncyCastle (the class names are shaded, but the `BC` provider name is taken), which is now deprecated and throws exceptions when used  - see [this blog post](https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html) for details.
2. `BCJcaJceHelper` (used internally by classes like `BaseWrapCipher`) [always looks at the system `BC` provider first](https://github.com/bcgit/bc-java/blob/master/prov/src/main/java/org/bouncycastle/jcajce/util/BCJcaJceHelper.java#L18), which means that it will use the deprecated default `BC` provider on Android.

This can be worked around by modifying the list of system security providers (using `System.removeProvider("BC")` and `System.addProvider(new BouncyCastleProvider())`), but I don't _think_ this should be necessary.

PhilGlass:
Closes #657.

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/util/BCJcaJceHelper.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/util/BCJcaJceHelper.java
@@ -15,9 +15,12 @@ public class BCJcaJceHelper
 
     private static synchronized Provider getBouncyCastleProvider()
     {
-        if (Security.getProvider("BC") != null)
+        final Provider system = Security.getProvider("BC");
+        // Avoid using the old, deprecated system BC provider on Android.
+        // See: https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html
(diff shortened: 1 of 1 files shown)


bcgit:
Merged. Thanks!

PR #677, opened 2020-03-17
Support for multiple armor headers in ArmoredOutputStream

vvvlado:
Currently ArmoredInputStream is able to handle multiple headers under the same name but ArmoredOutputStream is not able to create them. For example any additional "Comment" header overwrites the current one.

They are explicitly allowed by RFC 4880 in its section 6.2. Forming ASCII Armor.


bcgit:
Thanks for the patch. Merged!

PR #705, opened 2020-05-07
Replaced direct CommandMap usage with mailcap file - fixes issue #704

### LINKED ISSUE #704: bcmail not compatible with Karaf (OSGi)
The registration of CommandInfos in [SMIMESignedGenerator](https://github.com/bcgit/bc-java/blob/master/mail/src/main/java/org/bouncycastle/mail/smime/SMIMESignedGenerator.java#L123) produces problems.
In Apache Karaf this results in these commands being registered with the last OSGi bundle that was loaded with commands registered via a `META-INF/mailcap` file. With as result that the content handler class of those commands cannot be loaded due to the nature of OSGi class loaders.

Suggested fix is for bcmail to also make use of a `META-INF/mailcap` file. This ensures proper registration of the content handlers in an OSGi environment.

elmuerte:
Instead of directly updating the javax.activation CommandMap via code this change uses a mailcap file in the META-INF. This file is automatically processed by the javax.activation's standard MailcapCommandMap, and it is also processed by the special CommandMap implementation used by ServiceMix/Karaf.

This fixes all problems as described in #704

### DIFF
--- a/ant/bc+-build.xml
+++ b/ant/bc+-build.xml
@@ -536,16 +536,25 @@
              <fileset dir="${src.dir}">
                  <include name="org/bouncycastle/mail/**/*.java" />
                  <include name="org/bouncycastle/mail/**/*.properties" />
+                 <include name="META-INF/mailcap" />
                  <exclude name="**/*Test.java" />
                  <exclude name="**/*Tests.java" />
                  <exclude name="**/test/*.java" />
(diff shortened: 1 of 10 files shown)


elmuerte:
#704 is fixed with the changes in in 1.66b05; thus this PR is not needed anymore.

PR #722, opened 2020-06-07
Keep root cause of KeyStoreException exception

yschimke:
Long term fix for debugging https://github.com/square/okhttp/issues/6116

This particular error has come up a couple of times and without the root NPE, it's difficult to debug.


yschimke:
Is there a deeper reason to not keep that cause? Security, exposing internal structure, garbage collection/liveness issues?

dghgit:
Thanks for the report. I've added a way of passing back the exception. The reason for the way it's handled at the moment dates back to the days when there wasn't a nested exception constructor for KeyStore exception. Not a great reason I'll grant, but it's been quite a few years since there's been a need to look at this class.

PR #727, opened 2020-06-12
Fix DTLS handshake infinite loop

### LINKED ISSUE #726: DTLS handshake gets stuck with high latency between the client and the server
## Issue description
If there is a latency >= 500ms between the client and the server the handshake won't complete.

## How to reproduce it
I am experimenting with the example classes ```DTLSServerTest``` and ```DTLSClientTest``` under the package ```org.bouncycastle.tls.test```.
Before running the test I've set a latency of 500ms on my loopback interface with ```sudo tc qdisc add dev lo root netem delay 500ms```.

## Cause description
With this configuration, the __client receives the ```Hello Verify Request``` after its resend timeout (initially set to 1000ms) expires__, so it sends another ```Client Hello``` before it can receive the aforementioned ```Hello Verify Request```.

This causes an infinite loop where the client will continuously send (every time its resend timeout expires) the patched ```Client Hello + Cookie``` ignoring every response coming from the server.
On the other side, the server will continuously receive the ```Client Hello + Cookie``` replying with the ```Server Hello``` sequence (```Certificate```, ```ServerKeyExchange```, ```CertificateRequest```, ```ServerHelloDone```) that will be ignored by the client.

NicFontana:
Fix DTLS handshake infinite loop caused by the client dropping packets different from Client Hello kind of packets.

This happens when it receives the Hello Verify Request after its resend timeout (initially set to 1000ms) expires.

Fix #726

### DIFF
--- a/tls/src/main/java/org/bouncycastle/tls/DTLSRecordLayer.java
+++ b/tls/src/main/java/org/bouncycastle/tls/DTLSRecordLayer.java
@@ -559,7 +559,20 @@ class DTLSRecordLayer
 
             if (!isClientHelloFragment)
             {
-                return -1;
+                boolean isServerHelloFragment =
+                        getReadEpoch () == 0
+                                &&  length > 0
+                                &&  ContentType.handshake == recordType
+                                &&  HandshakeType.server_hello == TlsUtils.readUint8(record, RECORD_HEADER_LENGTH);
+
+                if (isServerHelloFragment)
+                {
+                    setReadVersion(recordVersion);
+                }
+                else
+                {
+                    return -1;
+                }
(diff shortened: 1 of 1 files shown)


peterdettman:
I think you've correctly identified `readVersion` handling as a problem, but it might be more straightforward to just not set it from HelloVerifyRequest `recordVersion`. The current code clears  `readVersion` after it receives a HelloVerifyRequest message, but that clearing doesn't happen for retransmitted records.

peterdettman:
I've fixed this by not setting `readVersion` from HelloVerifyRequest instead, but we have credited you with the initial fix.

PR #734, opened 2020-06-19
Fix padding error in CSHAKEDigest,java

mneuhaus-cv:
The current implementation og diffPadAndAbsorb contradicts the specification of bytepad in NIST SP 800-38.

Letting bit-padding aside, SP 800-185 specifies that...
3. while (len(z)/8) mod w ≠ 0:
z = z || 00000000

* z in SP800-185 is named diff in BCs implementation
* len(z) is the BITlength of z - len(z) /8 matches diff.length in BC implementation
* w is the padding width (in bytes) - this matches blockSize in the BC implementation.

When by chance z / diff has a length of blockSize, the current implementation will append an entire block (w) of ZEROs.
This contradicts the specification cited above.

Thanks for your work on BC!

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/digests/CSHAKEDigest.java
+++ b/core/src/main/java/org/bouncycastle/crypto/digests/CSHAKEDigest.java
@@ -40,6 +40,8 @@ public class CSHAKEDigest
 
         int required = blockSize - (diff.length % blockSize);
 
+        required %= blockSize;  //  otherwise diffPadAndAbsorb will add one entire blockSize of ZEROs,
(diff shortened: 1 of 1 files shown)


dghgit:
Thanks for the report. Yes, I checked, you are definitely correct. Very subtle one. This is now fixed in the latest beta, I went a slightly different route but also generated some test vectors for detecting the particular case. If you're interested in those they are in CSHAKETest.checkZeroPadZ()

Beta is now up at https://www.bouncycastle.org/betas 166b13 or later.

dghgit:
Fixed in 1.66 release. Thanks for again for the patch.

PR #737, opened 2020-06-22
support HMAC-SM3

Gsealy:
support for HMAC-SM3

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/digest/SM3.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/digest/SM3.java
@@ -1,8 +1,12 @@
 package org.bouncycastle.jcajce.provider.digest;
 
 import org.bouncycastle.asn1.gm.GMObjectIdentifiers;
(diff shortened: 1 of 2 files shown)


dghgit:
Thanks for the patch, now merged with minor changes. See https://www.bouncycastle.org/betas 166b13 or later.

Gsealy:
Thanks

prov/src/main/java/org/bouncycastle/jcajce/provider/digest/SM3.java
```diff
@@ -29,6 +33,27 @@ public Object clone()
         }
     }
 
+    /**
+     * SHA1 HMac
```
Gsealy, line 37: @dghgit sorry.. I was forget change javadoc, plz help me modify this, tks! 😃

dghgit: I guess I forgot to read it! Should be updated soon.

PR #745, opened 2020-07-08
Fix for NegativeArraySizeException in RFC5649WrapEngine

simon-greatrix:


core/src/main/java/org/bouncycastle/crypto/engines/RFC5649WrapEngine.java
```diff
@@ -154,7 +154,7 @@ public String getAlgorithmName()
             throw new InvalidCipherTextException("unwrap data must be a multiple of 8 bytes");
         }
 
-        if (n == 1)
+        if (n <= 1)
```
simon-greatrix, line 157: If the length was zero, then the it was processed as if it was 16 bytes, which caused a runtime exception.

core/src/main/java/org/bouncycastle/crypto/engines/RFC5649WrapEngine.java
```diff
@@ -217,7 +217,7 @@ public String getAlgorithmName()
 
         // Check the number of padded zeros
         int expectedZeros = upperBound - mli;
-        if (expectedZeros >= paddedPlaintext.length)
+        if (expectedZeros >= paddedPlaintext.length || expectedZeros<0)
```
simon-greatrix, line 220: Added catch for the expected number of zeros being negative.

core/src/test/java/org/bouncycastle/crypto/test/AESWrapPadTest.java
```diff
@@ -162,6 +163,58 @@ public void performTest()
             rnd.nextBytes(keyToWrap);
             wrapAndUnwrap(kek, keyToWrap);
         }
+
```
simon-greatrix, line 166: All the existing tests checked correct behaviour. I've added tests for incorrect behaviour.

core/src/main/java/org/bouncycastle/crypto/engines/RFC5649WrapEngine.java
```diff
@@ -215,12 +215,13 @@ public String getAlgorithmName()
...
-        // Check the number of padded zeros
+        // Check the number of padding zeros
         int expectedZeros = upperBound - mli;
-        if (expectedZeros >= paddedPlaintext.length)
+        if (expectedZeros >= 8 || expectedZeros < 0)
         {
+            // We have to pick a "typical" amount of padding to avoid timing attacks.
             isValid = false;
-            expectedZeros = paddedPlaintext.length;
+            expectedZeros = 4;
```
simon-greatrix, line 224: I believe this was a timing attack. The expected number of padding zeros was from 0 to 7. By performing a verification of the entire plain text we were doing a lot more work on the comparison than is typical.

dghgit:
Thanks for the patch! Now merged.

PR #746, opened 2020-07-13
Fix clearSignedFileProcessor header description.

macknight:
The code is written as 'System.err.println("usage:ClearSignedFileProcessor [-s file keyfile passPhrase]|[-v sigFile keyFile]");'
So fix the description to conform to the code.


dghgit:
Thanks for the patch. Now merged. Will show up in the next github sync (around 24 hours).

PR #756, opened 2020-07-23
Fixing typos from "lenght" to "length"

hablutzel1:


dghgit:
Thanks for the patch. This is now merged and should show up on github at the next sync (probably Monday).

dghgit:
Thanks for the patch. Now merged, should show up in the next github sync (around 24 hours or so).

PR #772, opened 2020-08-28
char[] to byte[] conversion

alek-sun:
Conversion without specifying an encoding cause errors when using non-ascii symbols in passwords.


dghgit:
The reason for not using UTF8 is historical as originally OpenSSL did not support UTF8. 

I've merged this for 1.67, hoping that there are enough UTF8 based versions of the PBE algorithm to have tilted the balance in its favor. It should appear in the next beta 167b06 on https://www.bouncycastle.org/betas

Issue #774, opened 2020-08-29
SHA224 is not supported with clear Sign - getting unspported hash algorithm


sreedharj:
Hi,

I am using ClearSignedFiledProcessor class to sign a file using Hash Algorithm  SHA224  ( HashAlgorithmTags.SHA224: ),  and getting the following  error.
java.io.IOException: unknown hash algorithm tag in beginClearText: 11
	at org.bouncycastle.bcpg.ArmoredOutputStream.beginClearText(Unknown Source)
	at com.b2b.pgp.security.ClearSignedFileProcessor.signFile(ClearSignedFileProcessor.java:285)

Looking at the ArmoredOutputStream  source,  I don't see support for SHA224.  Any reasons?

 */
    public void beginClearText(
        int    hashAlgorithm)
        throws IOException
    {
        String    hash;

        switch (hashAlgorithm)
        {
        case HashAlgorithmTags.SHA1:
            hash = "SHA1";
            break;
        case HashAlgorithmTags.SHA256:
            hash = "SHA256";
            break;
        case HashAlgorithmTags.SHA384:
            hash = "SHA384";
            break;
        case HashAlgorithmTags.SHA512:
            hash = "SHA512";
            break;
        case HashAlgorithmTags.MD2:
            hash = "MD2";
            break;
        case HashAlgorithmTags.MD5:
            hash = "MD5";
            break;
        case HashAlgorithmTags.RIPEMD160:
            hash = "RIPEMD160";
            break;
        default:
            throw new IOException("unknown hash algorithm tag in beginClearText: " + hashAlgorithm);
        }

dghgit:
It wasn't part of the original standard. This has been merged and will be in 167b06 on https://www.bouncycastle.org/betas

PR #775, opened 2020-08-29
Support for SHA 224 algorithm

Susmit07:
Support for SHA 224 Algorithm for Clear Text

Will fix issue - SHA224 is not supported with clear Sign - getting unspported hash algorithm #774


Susmit07:
@peterdettman can you please review once, and provide your feedback, as customers are willing to use SHA 224 as well, and somehow it was missed in case of ClearSignature

dghgit:

PR #782, opened 2020-09-10
Use BigIntegers.intValueExact() for compatibility

hvisser:
`BigInteger.intValueExact()` is only available since Java 8. This leads to run time errors on older JDKs (when using the 1.5on artifact) and on Android where this method is missing entirely.


peterdettman:
Thanks for the patch. Merged with some followup changes (mirroring here can take a few days).

peterdettman:
Fixed in 1.67.

PR #785, opened 2020-09-17
ESTService - Add Server Side Key Generation support. https://tools.ietf.org/html/r…

mkarasik:
ESTService
Add Server Side Key Generation support. https://tools.ietf.org/html/rfc7030#section-4.4


dghgit:
Apologies for the long silence on this one, but this was merged with revision, and Server Side Key Generation support was added to EST a few releases ago. Closing this off dropped off the radar. Thanks for the contribution.

Issue #773, opened 2020-08-28
No way to create BCEdDSAPublicKey from raw key data


cyberphone:
If you only have the public key value and curve name which is the case when using https://tools.ietf.org/html/rfc8037 it is unclear to me how to proceed.  None of the "spec" methods seem to fit and the BC tests do not include such transformations either.

```json
{
  "kty":"OKP",
  "crv":"Ed25519",
  "x":"11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo"
}
```

dghgit:
Try this: 

   private static X509EncodedKeySpec createEdDSAPublicKeySpec(byte[] rawPoint)
        throws IOException
    {
        AlgorithmIdentifier algId;
        if (rawPoint.length == 32)
        {
            algId = new AlgorithmIdentifier(EdECObjectIdentifiers.id_Ed25519);
        }
        else
        {
            algId = new AlgorithmIdentifier(EdECObjectIdentifiers.id_Ed448);
        }

        return new X509EncodedKeySpec(new SubjectPublicKeyInfo(algId, rawPoint).getEncoded());
    }

cyberphone:
Thanx!

PR #796, opened 2020-10-12
EdDSA: Add public constructors for key classes

sundbry:
Hello Lords of the Bouncy Castle,

Please consider this proposal to make these constructors public. We are relying on access to these constructors in the `buddy` library to generate EdDSA signatures and JSON web keys. It is important that we are able to construct an EdDSA key pair directly from its raw bytes.

https://github.com/funcool/buddy-core/pull/73


dghgit:
Sorry, can't do this. They're internal - if you move to the FIPS module you won't be able to find them.
Have a look at: #773 that sounds like it will show you how to deal with public keys. PrivateKeys are similar, you just need to use a PrivateKeyInfo and wrap the raw bytes in a DEROctetString().

PR #804, opened 2020-10-18
Test Vectors for EC Point Multiply

BlackthornYugen:
Test multiply with the test vectors downloaded from:
http://point-at-infinity.org/ecc/nisttv


peterdettman:
Merged with a few followup changes. Thanks for the PR.

PR #810, opened 2020-10-21
Removed redundant fields and null-checks

artem-smotrakov:
This is a little cleanup. The patch fixes some issues reported by LGTM

- Unused fields
- Redundant null-checks

https://lgtm.com/projects/g/bcgit/bc-java/?mode=list&tag=&id=java%2Funused-container%2Cjava%2Fuseless-null-check


dghgit:
Thanks for the patch! Merged with minor revisions.

PR #814, opened 2020-10-22
Suppress index-out-of-bound false-positives from LGTM

artem-smotrakov:
LGTM reports several findings about accessing an array with an index that is greater than or equal to the length of the array. Those are false-positives.

https://lgtm.com/projects/g/bcgit/bc-java/?mode=list&tag=&id=java%2Findex-out-of-bounds


dghgit:
Thanks for the patch. Merged.

PR #818, opened 2020-10-23
Enable nohttp check during the build

artem-smotrakov:
HTTP is a plaintext protocol which means that someone may be able to eavesdrop the data. To prevent this, HTTPS should be used whenever possible. However, maintaining using `https://` in all URLs may be difficult. The `nohttp` tool can help here. The tool scans all the files in a repository and reports where `http://` is used.

Here is a list of main updates:

- Updated `build.gradle` to call `nohttp` during the build.
- Updated `http://` to `https://` in URLs.
- Added an exclude list.

Sorry for many updated files. All the findings had to be fixed or suppressed.

Here are some info about the tool:

- https://github.com/spring-io/nohttp
- https://spring.io/blog/2019/06/10/announcing-nohttp

### DIFF
--- a/CONTRIBUTORS.html
+++ b/CONTRIBUTORS.html
@@ -78,18 +78,18 @@ Organisations
 </p>
 <ul>
 <li>Holders of <a href="https://www.cryptoworkshop.com">Crypto Workshop Support Contracts</a>. Without the consulting time left over from support contracts being contributed back to working on the Bouncy Castle APIs, progress would be impossible. You know who you are!</li>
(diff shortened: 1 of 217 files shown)


dghgit:
I'm working through this one, just have to be a little careful in case there is something which is actually supposed to be http.

dghgit:
Thanks for the patch, I think this has now been done in all the places it is appropriate to do so.

artem-smotrakov:
> Thanks for the patch, I think this has now been done in all the places it is appropriate to do so.

@dghgit The purpose of this tool is to prevent unsafe http:// link from appearing again.

dghgit (commented):
updated to use 0.0.8, thanks for providing the config files as well. It should appear on github in the next day.

PR #848, opened 2020-12-05
Allow population of PGPSignatureSubpacketGenerator from given PGPSign…

vanitasvitae:
…atureSubpacketVector

Hey :)
I'm struggling to extend the expiration date of a keyring while preserving the signature subpackets of the previous self-signature.
This PR allows for the `PGPSignatureSubpacketGenerator` to be prepopulated with subpackets of a given `PGPSignatureSubpacketVector`.
I also prevent some packet types from being added into the vector multiple times where it makes sense to prevent this.
This is being done by switching from an ominous list of subpackets to members for the different subpacket types.

The changes are aimed to be as backwards compatible as possible, so where the name of a method doesn't match what it does (the case with some setters) I deprecated the unsuitable method and forward method calls to a method that is more suitably named.

Let me know what you think of this change or if I can improve it somehow.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSignatureSubpacketGenerator.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSignatureSubpacketGenerator.java
@@ -33,25 +33,151 @@ import org.bouncycastle.bcpg.sig.TrustSignature;
  */
 public class PGPSignatureSubpacketGenerator
 {
-    List list = new ArrayList();
+    private SignatureCreationTime signatureCreationTime;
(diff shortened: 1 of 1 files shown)


vanitasvitae:
Regarding the now deprecated methods: Should I add another commit that changes calls to those methods from inside bouncycastle to instead call their replacements?

dghgit:
I've done a merge and edit on this. The basic idea is a good one, as are a lot of the method name changes - I think we're stuck with the list though as it allows people to define the order packets appear in, I think the principal thing missing is the ability to remove packets from the generator. The code should be appearing on github shortly. Note I also added hashCode() and equals() to SignatureSubpacket.

PR #859, opened 2020-12-27
Pull setWithIntegrityPacket up into PGPDataEncryptorBuilder interface

vanitasvitae:
Both the Jce as well as the Bc implementation of `PGPDataEncryptorBuilder` do have a `setWithIntegrityPacket(boolean)` method, which however is not present in the parent interface.

This PR simply pulls up the method declaration into the interface.


dghgit:
Thanks for the patch. Merged!

PR #875, opened 2021-02-03
create Kangaroo digests

tonywasher:
Please note that this is an implementation of Kangaroo12 and Marsupilami14. The latter appears in the [`base spec`](https://keccak.team/files/KangarooTwelve.pdf), but not in the [ietf draft](https://tools.ietf.org/html/draft-viguier-kangarootwelve-04). Test vectors are only available for KangarooTwelve, but Marsupilami14 is such a minor variant that I left it in.


dghgit:
Thanks for the update. Merged with minor revisions.

PR #876, opened 2021-02-04
Add Blake3 support

tonywasher:


jvz:
I have a reference implementation in Java [here](https://github.com/o1c-dev/o1c/tree/main/src/main/java/dev/o1c/modern/blake3) which can potentially be ported/copied here.

core/src/main/java/org/bouncycastle/crypto/digests/Blake3Digest.java
```diff
@@ -0,0 +1,907 @@
...
+
+    /**
+     * The indices.
+     */
+    private final byte[] theIndices = new byte[NUMWORDS << 1];
+
+    /**
+     * The chainingStack.
+     */
+    private final Deque<int[]> theStack = new ArrayDeque();
```
jvz, line 209: Nit: this could specify an initial deque size of 54 since that's the max size the stack gets. Or a fixed size array to avoid additional garbage during hashing.

core/src/main/java/org/bouncycastle/crypto/digests/Blake3Digest.java
```diff
@@ -0,0 +1,907 @@
...
+                      final int posC,
+                      final int posD)
+    {
+        /* Determine indices */
+        int msg = msgIdx << 1;
+        int rot = 0;
+
+        /* Perform the Round */
+        theV[posA] += theV[posB] + theM[theIndices[msg++]];
+        theV[posD] = rotr32(theV[posD] ^ theV[posA], ROTATE[rot++]);
```
jvz, line 736: Nit: this can use `Integer.rotateRight`, available since Java 1.5.

core/src/main/java/org/bouncycastle/crypto/params/Blake3Parameters.java
```diff
@@ -0,0 +1,84 @@
...
+     * The context.
+     */
+    private byte[] theContext;
+
+    /**
+     * Create a key parameter.
+     * @param pContext the context
+     * @return the parameter
+     */
+    public static Blake3Parameters context(final byte[] pContext)
```
jvz, line 32: It may be worth adding another factory method using a `String` parameter which encodes to UTF-8 as Blake3 encourages use of human-readable unique strings for the KDF context parameter (e.g., an ISO-8601 timestamp and application name).

jvz (approved):
Looks decent. I initially thought this was an issue and not a PR, so here's a free review.

core/src/main/java/org/bouncycastle/crypto/digests/Blake3Digest.java
dghgit: Thanks for the comment. At the moment it's better like this, an even bigger nit we have to deal with is we have people using Java 4... Sun licensed so many JVMs, it's quite a broad church still.

core/src/main/java/org/bouncycastle/crypto/params/Blake3Parameters.java
dghgit: Thanks, I'll note this for when it's added to the JCA provider (the next step).

dghgit:
Thanks, merged with minor changes. @jvz thanks also for additional review and comments.

PR #878, opened 2021-02-10
PEMParser improved for inheritance

jpstotz:
Especially that the central `parsers` HashMap was declared private makes it impossible to extend PEMParser and e.g. to add other proprietary PEM formats.

Additionally I added all types as constants and made it possible to get a list of supported types.  

In `readObject()` method access of the `parsers` was inefficient, two times a look-up where one is sufficient.


dghgit:
Thanks for the patch. I can see the merit in that... Merged.

PR #886, opened 2021-02-15
Properly rewind streams for messages with multiple SKESK packets

### LINKED ISSUE #880: OpenPGP: Encryption with multiple symmetric passwords sometimes fails.
Hey!
When I encrypt a message with two passphrases symmetrically, my code sometimes generates a PGP message that cannot be decrypted again.
This issue happens at around 3.5% of all encrypted messages.

I created a [test class](https://github.com/pgpainless/pgpainless/blob/bug72/pgpainless-core/src/test/java/org/bouncycastle/bug72/Bug72.java) to reproduce the issue.

My encryption method looks like this:
```
    /**
     * This method sometimes generates broken ciphertext.
     * @return
     * @throws IOException
     * @throws PGPException
     */
    public byte[] encryptionBouncycastle(byte[] plaintext, String[] passwords) throws IOException, PGPException {
        ByteArrayInputStream in = new ByteArrayInputStream(plaintext);
        ByteArrayOutputStream cipherText = new ByteArrayOutputStream();
        ArmoredOutputStream armorOutputStream = new ArmoredOutputStream(cipherText);

        BcPGPDataEncryptorBuilder dataEncryptorBuilder = new BcPGPDataEncryptorBuilder(SymmetricKeyAlgorithmTags.AES_256);
        dataEncryptorBuilder.setWithIntegrityPacket(true);

        PGPEncryptedDataGenerator encryptedDataGenerator = new PGPEncryptedDataGenerator(dataEncryptorBuilder);

        for (String password : passwords) {
            encryptedDataGenerator.addMethod(new BcPBEKeyEncryptionMethodGenerator(password.toCharArray()));
        }

        OutputStream encryptionStream = encryptedDataGenerator.open(armorOutputStream, new byte[BUFFER_SIZE]);

        PGPCompressedDataGenerator compressedDataGenerator = new PGPCompressedDataGenerator(CompressionAlgorithmTags.UNCOMPRESSED);
        OutputStream compressionStream = new BCPGOutputStream(compressedDataGenerator.open(encryptionStream));

        OutputStream literalDataStream = literalDataGenerator.open(compressionStream, PGPLiteralData.BINARY, PGPLiteralData.CONSOLE, new Date(), new byte[BUFFER_SIZE]);

        OutputStream myOutputStream = new MyOutputStream(literalDataStream, compressionStream, encryptionStream, armorOutputStream);
        Streams.pipeAll(in, myOutputStream);
        myOutputStream.close();

        return cipherText.toByteArray();
    }
```

An example of a broken message would be this:
```
-----BEGIN PGP MESSAGE-----
Version: BCPG v1.68

jC4ECQMC3HsxjIBGJ7lg9pXvZ93TZpD76Lv+JVxpi3DWKNGRh7QKqhE+FqnMA96n
jC4ECQMCLYvyr3wa8qlggqbncZuhOh6M4o3/d9UkeDHYRY5JwGQ9rfjmYhp45e7d
0kgBS4Xxaim+KHeavy36u/ohad0BkR8sQCWmAlSIfYT7yXR8sKj4zjqIT6/FqGcF
wZ9QgP/uwqVI/d/PC/1CCE2QaJ1Pp9XAQQw=
=9iqo
-----END PGP MESSAGE-----
```
It was created using the code above and encrypted for "password1" and "password2".
Decrypting it with "password2" fails in this case, while "password1" works.
GnuPG complains that the sessionkey is broken.

Some observation I made is that in some cases `pbeEncryptedData.getSymmetricAlgorithm(passphraseDecryptor)` would return the wrong symmetric algorithm tag (eg. Camellia-256 while AES-256 is being used. This would strengthen my encoding issue theory.

Is there an encoding issue happening?
Or did I simply made a mistake in my code?

vanitasvitae:
This fixes #880

When an OpenPGP message was encrypted with multiple encryption methods, eg. symmetric keys,
it would sometimes happen that the first SKESK (symmetric-key encrypted session key packet) would be decrypted with a wrong passphrase. This would most of the time fail, as the very first byte of the result that notates the used symmetric encryption algorithm would often not be a valid algorithm id.
However, sometimes (3.5% of times) it would match a valid algorithm ID.
Decryption would then commence, reading some bytes from the SEIP (symmetrically encrypted integrity-protected data packet).
The number of bytes that would be read is equal to the length of the IV of the symmetric encryption algorithm plus 2 bytes checksum.
Decryption would then finally fail due to checksum mismatch.
Note that BC would not rewind the SEIP inputstream.

Now when the next SKESK would be decrypted using the correct passphrase, BC would again decrypt it, this time resulting in the correct session key. When this session key was used to decrypt the SEIP inputstream, either an EOFException would be thrown, or the checksum would again mismatch, as the first (iv.length + 2) octets were lost in the previous decryption attempt.

This PR fixes the issue by wrapping the SEIP input stream with a bufferedInputStream and resetting it after a decryption attempt failed.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/InputStreamPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/InputStreamPacket.java
@@ -1,27 +1,27 @@
 package org.bouncycastle.bcpg;
 
+import java.io.BufferedInputStream;
+
 /**
  * A block of data associated with other packets in a PGP object stream.
  */
 public class InputStreamPacket
     extends Packet
 {
-    private BCPGInputStream        in;
+    private BufferedInputStream        in;
 
     public InputStreamPacket(
         BCPGInputStream  in)
     {
-        this.in = in;
+        this.in = new BufferedInputStream(in);
     }
 
     /**
      * Obtains an input stream to read the contents of the packet.
-     * <p>
-     * Note: you can only read from this once...
-     * </p>
+     *
      * @return the data in this packet.
      */
-    public BCPGInputStream getInputStream()
+    public BufferedInputStream getInputStream()
     {
         return in;
     }
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPPBEEncryptedData.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPPBEEncryptedData.java
@@ -1,5 +1,6 @@
 package org.bouncycastle.openpgp;
 
+import java.io.BufferedInputStream;
 import java.io.EOFException;
 import java.io.InputStream;
 
@@ -84,7 +85,9 @@ public class PGPPBEEncryptedData
 
             PGPDataDecryptor dataDecryptor = dataDecryptorFactory.createDataDecryptor(withIntegrityPacket, sessionData[0] & 0xff, sessionKey);
 
-            encStream = new BCPGInputStream(dataDecryptor.getInputStream(encData.getInputStream()));
+            BufferedInputStream bufferedIn = encData.getInputStream();
+            bufferedIn.mark(dataDecryptor.getBlockSize() + 2); // iv + 2 octets checksum
+            encStream = new BCPGInputStream(dataDecryptor.getInputStream(bufferedIn));
 
             if (withIntegrityPacket)
             {
@@ -102,6 +105,7 @@ public class PGPPBEEncryptedData
 
                 if (ch < 0)
                 {
+                    bufferedIn.reset();
                     throw new EOFException("unexpected end of stream.");
                 }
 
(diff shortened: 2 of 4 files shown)


vanitasvitae:
Actually this not only affects messages with multiple SKESK packets, but also messages with more than one PKESK packets when wildcard recipient IDs were used.

dghgit:
Merged with some changes. I think I managed to avoid the API change. The patch for the PublicKey data didn't actually seem to do anything (there's no checksum due to an earlier PGP issue, so any errors are in the actual decryption only get picked up at the end of the stream, there will be an exception earlier though if the session key checksum check fails). Pushed and up on https://www.bouncycastle.org/betas 169b04 or later.

PR #888, opened 2021-02-15
Set fixed length for DEROctetString representation of ed25519 keys to 32

### LINKED ISSUE #887: OpenPGP: EdDSA: private key encoding has incorrect length
Hey!
When I generate EdDSA based OpenPGP keys, I sometimes run into the issue that the generated private keys have wrong length.
The expected length is 32, but some private keys appear to have length 31 instead.

Example:
```
version: 0
privateKeyAlgorithm: 1.3.101.112
private key: #041f973cef5e047bdf6ee18e821252369fe6b7e4de83bca98f032b31a4d5abb2e4
```

The key was generated using this code (repeat 1000 times or so to reproduce):
```java
        Date creationDate = new Date();
        int algorithm = PublicKeyAlgorithmTags.EDDSA;
        KeyPairGenerator certKeyGenerator = KeyPairGenerator.getInstance("ed25519", new BouncyCastleProvider());
        certKeyGenerator.initialize(new ECNamedCurveGenParameterSpec("ed25519"));
        KeyPair keyPair = certKeyGenerator.generateKeyPair();

        BcPGPKeyConverter converter = new BcPGPKeyConverter();
        PGPKeyPair jcaPgpPair = new JcaPGPKeyPair(algorithm, keyPair, new Date());
        AsymmetricKeyParameter publicKey = converter.getPublicKey(jcaPgpPair.getPublicKey());
        AsymmetricKeyParameter privateKey = converter.getPrivateKey(jcaPgpPair.getPrivateKey()); // This line throws.
        AsymmetricCipherKeyPair asymKeyPair = new AsymmetricCipherKeyPair(publicKey, privateKey);

        PGPKeyPair bcKeyPair = new BcPGPKeyPair(algorithm, asymKeyPair, new Date());
```

As a side-question: Is there a better way to get from the KeyPair to the BcPGPKeyPair without the detour via the JcaPGPKeyPair?
My goal is to separate JcaJce and Bc implementations, but so far using the JcaPGPKeyPair was the only way I found to generate OpenPGP keys with Bc.

What could cause this issue?

vanitasvitae:
This seems to fix #887 
The issue was that sometimes the byte array representation of the X component of the ed25519 private key would not be of length 32 (eg. 31).
This PR limits the length of the byte array to exactly 32.

I'm not sure if this same fix can be applied to other algorithms as well. This might be something for you worth looking into :)

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcPGPKeyConverter.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcPGPKeyConverter.java
@@ -135,7 +135,7 @@ public class BcPGPKeyConverter
 
                 return implGetPrivateKeyPKCS8(new PrivateKeyInfo(
                     new AlgorithmIdentifier(EdECObjectIdentifiers.id_Ed25519),
-                    new DEROctetString(BigIntegers.asUnsignedByteArray(eddsaK.getX()))));
+                    new DEROctetString(BigIntegers.asUnsignedByteArray(32, eddsaK.getX()))));
             }
 
             case PublicKeyAlgorithmTags.ELGAMAL_ENCRYPT:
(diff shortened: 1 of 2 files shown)


dghgit:

PR #889, opened 2021-02-17
ArmoredOutputStream: Add method to clear out all headers

vanitasvitae:
This PR adds a method to remove all headers from the ArmoredOutputStream.
This is useful to prevent fingerprinting.


dghgit:
I've added this - one small change method is no called "clearHeaders()". Thanks.

PR #891, opened 2021-02-19
Skip marker packets when parsing public key rings

vanitasvitae:
According to [rfc4880](https://tools.ietf.org/html/rfc4880#section-5.8) marker packets are obsolete and need to be ignored when encountered.

If the user tried to parse a public key ring which was prefixed with a marker packet, parsing would fail as the inital tag would match neither public key packet nor subkey packet.

My fix skips any marker packets and continues to parse the public key ring after skipping.

This fixes parts of https://github.com/pgpainless/pgpainless/issues/84


dghgit:
I guess there is a first time for everything. Thanks for the patch, merged with minor revision.

PR #895, opened 2021-03-01
Allow empty password in PKCS12KeyStoreSpi.

Yqwed:
Previously empty array was considered as effective password.


dghgit:
Thanks for the patch. Merged.

PR #902, opened 2021-03-10
Update ESTService.java

dipakbag:
getCACerts not able to retrieve cacerts from OpenXPKI ca server through est , it throws ESTException . In OpenXPKI ca server response the content-type as "application/pkcs7-mime; smime-type=certs-only".


dghgit:
Merged with minor revisions.

PR #904, opened 2021-03-12
Generify most lists and iterators in PGP key/-ring classes

vanitasvitae:
This PR adds generics to most lists and iterator methods of `PGPPublicKey`, `PGPSecretKey`, `PGPPublicKeyRing`, `PGPSecretKeyRing`, `PGPKeyRing`.

It also fixes some possible `ConcurrentModificatonExceptions` when removing from a list by iterating over it with increasing index.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPKeyRingGenerator.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPKeyRingGenerator.java
@@ -18,10 +18,10 @@ import org.bouncycastle.openpgp.operator.PGPDigestCalculator;
  */
 public class PGPKeyRingGenerator
 {    
-    List                                keys = new ArrayList();
+    List<PGPSecretKey>                  keys = new ArrayList<>();
 
     private PBESecretKeyEncryptor       keyEncryptor;
-    private PGPDigestCalculator checksumCalculator;
+    private PGPDigestCalculator         checksumCalculator;
     private PGPKeyPair                  masterKey;
     private PGPSignatureSubpacketVector hashedPcks;
     private PGPSignatureSubpacketVector unhashedPcks;
@@ -87,8 +87,8 @@ public class PGPKeyRingGenerator
         this.checksumCalculator = checksumCalculator;
         this.keySignerBuilder = keySignerBuilder;
 
-        PGPSignature certSig = (PGPSignature)originalSecretRing.getPublicKey().getSignatures().next();
-        List hashedVec = new ArrayList();
+        PGPSignature certSig = originalSecretRing.getPublicKey().getSignatures().next();
+        List<SignatureSubpacket> hashedVec = new ArrayList<>();
         PGPSignatureSubpacketVector existing = certSig.getHashedSubPackets();
         for (int i = 0; i != existing.size(); i++)
         {
@@ -99,7 +99,7 @@ public class PGPKeyRingGenerator
             hashedVec.add(existing.packets[i]);
         }
         this.hashedPcks = new PGPSignatureSubpacketVector(
-            (SignatureSubpacket[])hashedVec.toArray(new SignatureSubpacket[hashedVec.size()]));
+                hashedVec.toArray(new SignatureSubpacket[0]));
         this.unhashedPcks = certSig.getUnhashedSubPackets();
 
         keys.addAll(originalSecretRing.keys);
@@ -146,7 +146,7 @@ public class PGPKeyRingGenerator
             sGen.setHashedSubpackets(hashedPcks);
             sGen.setUnhashedSubpackets(unhashedPcks);
 
-            List                 subSigs = new ArrayList();
+            List<PGPSignature> subSigs = new ArrayList<>();
             
             subSigs.add(sGen.generateCertification(masterKey.getPublicKey(), keyPair.getPublicKey()));
 
@@ -184,14 +184,14 @@ public class PGPKeyRingGenerator
      */
     public PGPPublicKeyRing generatePublicKeyRing()
     {
-        Iterator it = keys.iterator();
-        List     pubKeys = new ArrayList();
+        Iterator<PGPSecretKey> it = keys.iterator();
+        List<PGPPublicKey>     pubKeys = new ArrayList<>();
         
-        pubKeys.add(((PGPSecretKey)it.next()).getPublicKey());
+        pubKeys.add(it.next().getPublicKey());
         
         while (it.hasNext())
         {
-            pubKeys.add(((PGPSecretKey)it.next()).getPublicKey());
+            pubKeys.add(it.next().getPublicKey());
         }
         
         return new PGPPublicKeyRing(pubKeys);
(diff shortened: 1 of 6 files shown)


pg/src/main/java/org/bouncycastle/openpgp/PGPKeyRingGenerator.java
```diff
@@ -99,7 +99,7 @@ public PGPKeyRingGenerator(
             hashedVec.add(existing.packets[i]);
         }
         this.hashedPcks = new PGPSignatureSubpacketVector(
-            (SignatureSubpacket[])hashedVec.toArray(new SignatureSubpacket[hashedVec.size()]));
+                hashedVec.toArray(new SignatureSubpacket[0]));
```
vanitasvitae, line 102: Here we can omit `hashedVec.size()`: https://stackoverflow.com/a/29444594

pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKey.java
```diff
@@ -921,7 +914,7 @@ private static PGPPublicKey removeCert(
         PGPPublicKey    returnKey = new PGPPublicKey(key);
         boolean         found = false;
 
-        for (int i = 0; i < returnKey.ids.size(); i++)
```
vanitasvitae, line 924: Iterating the list forwards and removing from it can cause `ConcurrentModificationExceptions`. Therefore I reversed the direction.

pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKeyRing.java
```diff
@@ -338,12 +338,12 @@ public static PGPPublicKeyRing removePublicKey(
         PGPPublicKeyRing  pubRing,
         PGPPublicKey      pubKey)
     {
-        List       keys = new ArrayList(pubRing.keys);
+        List<PGPPublicKey> keys = new ArrayList<>(pubRing.keys);
         boolean    found = false;
         
-        for (int i = 0; i < keys.size();i++)
```
vanitasvitae, line 344: Again, I reversed direction of the for-loop to prevent ConcurrentModificationExceptions.

pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKeyRing.java
```diff
@@ -190,7 +190,7 @@ public PGPPublicKey getPublicKey(
 
         for (int i = 0; i != extraPubKeys.size(); i++)
         {
-            PGPPublicKey    k = (PGPPublicKey)keys.get(i);
```
vanitasvitae, line 193: Here I guess the code accidentally accessed the wrong list.

pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKeyRing.java
```diff
@@ -218,7 +218,7 @@ public PGPPublicKey getPublicKey(byte[] fingerprint)
...
```
vanitasvitae, line 221: Same here.

pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKeyRing.java
```diff
@@ -505,12 +505,12 @@ public static PGPSecretKeyRing removeSecretKey(
         PGPSecretKeyRing  secRing,
         PGPSecretKey      secKey)
     {
-        List       keys = new ArrayList(secRing.keys);
+        List<PGPSecretKey>       keys = new ArrayList<>(secRing.keys);
         boolean    found = false;
         
-        for (int i = 0; i < keys.size();i++)
```
vanitasvitae, line 511: Once again, I reversed the order of the loop.

vanitasvitae (commented):
I added a review with comments that explain some of the changes.

pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKeyRing.java
dghgit: Ooops... Thanks, Applied!

pg/src/main/java/org/bouncycastle/openpgp/PGPSecretKeyRing.java
dghgit: Thanks. Applied.

pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKeyRing.java
dghgit: The operation is on an internal list in an immutable class (well at least it's supposed to be). I think this might be a step further than necessary.

pg/src/main/java/org/bouncycastle/openpgp/PGPKeyRingGenerator.java
dghgit: Yes, interesting... I don't know, most of the PGP API is the same classes right down to Java 1. It is heavily used on Java 4 and Java 3 platforms (the new "j2me") as well. We filter out the generics for Java 4 and below, but the "meaningless" casts are still required for the other profiles based on the older JVMs. I'm not really sure which variant serves the greater good in this case.

vanitasvitae: the array initialization is just a micro-optimization anyways.

>but the "meaningless" casts are still required for the other profiles based on the older JVMs

That's interesting, I did not know that.

dghgit: Yes, it's a very broad church around here. I'm working my way through looking for public methods we've missed.

pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKeyRing.java
vanitasvitae: Shall I update my PR removing those changes, or do you just pick whatever changes you like?

dghgit: I'm going okay just picking things up I think. I'll close this one when I think I'm done.

dghgit:
Okay, think I'm done. I'm hoping you'll find that all the public API use of generics are now in place. I can still build the API using Java 4 as well, so I think everyone should be happy.... If I've missed anything let me know should sync up shortly.

vanitasvitae:
Are you plan to include the changes to `PGPPublicKey` (e.g. `public Iterator<PGPSignature> getSignaturesOfType(int signatureType)` as well?

dghgit:
That should be there now as well. Any others missed?

vanitasvitae:
The last commit I can see is from 23h ago. Did you forget to push the changes to `PGPPublicKey`?

dghgit:
Pushed but not synced to the mirror. I've just done a sync.

vanitasvitae:
Very nice!
One change that's still missing is added generics to `PGPPublicKey.getSignaturesForID(UserIDPacket id)`, otherwise it looks good :+1: :)

dghgit:
Done! Thanks!

vanitasvitae:
Thank YOU ;)

PR #906, opened 2021-03-13
Fix encoding of AlgorithmIdentifier.parameters of Digests, add support of Shake-based signatures.

xipki:
For the SHAKE-based signatures, I cannot find test vectors, so please review the code before merge.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
@@ -240,6 +240,10 @@ public class ArmoredInputStream
                     {
                         break;
                     }
+                    if (headerList.size() != 0 && line.indexOf(':') < 0)
+                    {
+                        throw new IOException("invalid armor header");
+                    }
                     headerList.add(line);
                     buf.setLength(0);
                 }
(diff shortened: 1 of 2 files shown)


dghgit:
I think we'll need to take small steps with this one, I agree with most of it, but it's a bit of a rabbit warren. For example the change:
       
else if (sigAlgId.getAlgorithm().equals(EdECObjectIdentifiers.id_Ed448))
        {
-      digAlgId = new AlgorithmIdentifier(NISTObjectIdentifiers.id_shake256_len, new ASN1Integer(512));
+            digAlgOid = NISTObjectIdentifiers.id_shake256;
        }

is incorrect. What's currently there is based on RFC 8419, Section 3.1. It's very easy to blow one's foot off here... If there's no objection, I'll make some small corrections first and then get you to update.

dghgit:
One other one, would you generate a couple of PSS signatures using SHAKE128 and SHAKE256 and send me enough details so I can verify them. There's another way of adding PSS for these, assuming we both produce the same results there is a better chance that the RFC has been followed correctly. Thanks.

xipki:
I will re-submit the requests.

dghgit:
Sure, by the way I've found a source of test vectors for PSS.

dghgit:
I've just pushed some code for PSS and ECDSA, it combines RFC 8702 and the draft of FIPS 186-5. Please take a look and poke a bit... 

The only doubtful one seems to be ECDSA with SHAKE128 not passing the test vectors I have, but they may actually be incorrect. The SHAKE256 vectors from the same provider do match up, investigating.

xipki:
Reviewed your changes, looks very well.

It would be better, if you can also specify the source of the test vectors.

dghgit:
Thanks, I generated the vectors from the ACVP demo system. They are apparently planning to start publishing vectors for download sometime soon - when that happens I'll update so people have a definitive source to check against.

PR #923, opened 2021-04-04
Expose PGPSignature.getDigestPrefix() to get fingerprint of signature

vanitasvitae:
This PR adds `PGPSignature.getDigestPrefix()` which calls `SignaturePacket.getFingerPrint()` to expose the signatures digest  (the left 16 bits of the signed hash value) to the user.


dghgit:
Thanks for the patch, merged. Will appear in 169b05 (and later) on https://www.bouncycastle.org/betas

PR #924, opened 2021-04-04
Update PKIXCRLUtil.java

pelzvieh:
This fixes a NullPointerException in findCRLs, when a CRL does not have the nextUpdate attribute set.


dghgit:
Thanks for the patch, merged. Will appear in 169b05 (and later) on https://www.bouncycastle.org/betas

PR #930, opened 2021-04-15
RFC 5280 boundaries checks

bjpbakker:
Checks the boundaries specified in RFC 5280 for `CRLReason`, `CRLNumber` and `IssuerSerial`.

`CRLReason` must be an integer between 0 and 10. `CRLNumber` and `IssuerSerial` must have a maximum size of 20 octets.

### DIFF
--- a/core/src/main/java/org/bouncycastle/asn1/x509/CRLNumber.java
+++ b/core/src/main/java/org/bouncycastle/asn1/x509/CRLNumber.java
@@ -20,6 +20,9 @@ public class CRLNumber
     public CRLNumber(
         BigInteger number)
     {
+        if (BigInteger.ZERO.compareTo(number) > 0) {
+            throw new IllegalArgumentException("Invalid CRL number : not in (0..MAX)");
+        }
         this.number = number;
     }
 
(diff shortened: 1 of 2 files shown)


xipki:
In my opinion, these three fixes are not correct. Details as follows:
1. CRLReason: RFC 5280 specifies only value from 0 to 10, but this may be extended,e.g. other specification or private usage. 
2. CRLNumber: The creator should use max. 160 bit CRL number, but this restriction does not applied to the receiver. 
3. IssuerSerial: The restriction on SerialNumber is the same as the CRLNumber as mentioned above.

ties:
Other values are specified: ISO/IEC 9594-8:2020 defines reason `weakAlgorithmOrKey (11)`. Furthermore it seems to be more lenient than RFC 5280: `CRLNumber ::= INTEGER(0..MAX)`, `IssuerSerial` re-uses `CertificateSerialNumber` which is `CertificateSerialNumber ::= INTEGER`.

It boils down to what the library is meant to be used for: Web or generic X.509.

bjpbakker:
Given that BC is a general purpose library (i.e. not web specific), I've updated these checks to only check the lower bounds for `CRLReason` and for `CRLNumber`.

Re-reading RFC 5280 on Issuer Serial Number I also dropped the non-negative integer check, as it states that:

    Non-conforming CAs may issue certificates with serial numbers
    that are negative or zero.  Certificate users SHOULD be prepared to
    gracefully handle such certificates.

These changes are also rebased on latest master now.

dghgit:
Thanks for the patch, also thanks to those who pointed out the original issues. Final result now merged.

PR #932, opened 2021-04-16
Add binary-signing-time CMS attribute [RFC6019]

bjpbakker:
Include the binary-sigining-time CMS attribute as described in RFC 6019.


dghgit:

PR #936, opened 2021-04-22
Extract dedicated functions out of SignerInformation.doVerify

ties:
Refactoring due to a security review of our code which ended up commenting on this part of BC:
In a report an comment was made about `SignerInformation.doVerify`. The author
suggested to split out the control paths to add a number of functions that have
a single responsibility.

This commit is a mechanical refactoring which extracts part of doVerify to separate functions.


mwcw:
Thanks,

This has been merged and will be synced later.

MW

PR #937, opened 2021-04-23
New implementation of BasicConstraintsValidation

MatthiasValvekens:
As discussed over e-mail earlier this week, here's a pull request with my new implementation of the `BasicConstraintsValidation` class in the `org.bouncycastle.cert.path.validations package`.

The new implementation fixes some bugs in the previous one, adds a check for self-issued certificates and adds regression tests from PKITS. In terms of public API it should be fully backwards-compatible with the previous implementation.

The test class I added contains all BasicConstraints tests from PKITS. I just copied the PKITS test data from the `prov` module into a folder under the test resource directory for the `pkix` module. I figured that this kind of duplication wouldn't appreciably inflate the size of the git repository, and it allowed me to use Java's resource loading mechanism to read the test data. If you'd prefer handling the test data provisioning in another way, or if there's anything else you'd like me to amend, please let me know!


dghgit:
Thanks for the patch. Merged with minor compatibility changes.

PR #940, opened 2021-04-27
JcaPGPKeyConverter: Set fixed length for DEROctetString representatio…

vanitasvitae:
…n of ed25519 keys

While a similar change was applied to the BcPGPKeyConverter (#888), the same issue apparently also affects the JcaPGPKeyConverter


dghgit:
Thanks for the patch. Merged!

Issue #942, opened 2021-04-27
McElieceKobaraImaiCipher - Getting randomly "Bad Padding: invalid ciphertext" exception while decrypting


mcarleio:
First of, I am using JDK 11 and bcprov-ext-jdk15on in version 1.68.

When you execute below code, you will often pass 100 iterations before the following error is raised:

    Exception in thread "main" org.bouncycastle.crypto.InvalidCipherTextException: Bad Padding: invalid ciphertext
	at org.bouncycastle.pqc.crypto.mceliece.McElieceKobaraImaiCipher.messageDecrypt(Unknown Source)

Short explanation of the code:

1. generate random bytes
2. generate McEliece KeyPair
3. In a loop:
    1. Encrypt the bytes
    2. Decrypt the encrypted bytes

Here is the demo code:

    import org.bouncycastle.crypto.AsymmetricCipherKeyPair;
    import org.bouncycastle.crypto.params.AsymmetricKeyParameter;
    import org.bouncycastle.pqc.crypto.mceliece.McElieceCCA2KeyGenerationParameters;
    import org.bouncycastle.pqc.crypto.mceliece.McElieceCCA2KeyPairGenerator;
    import org.bouncycastle.pqc.crypto.mceliece.McElieceCCA2Parameters;
    import org.bouncycastle.pqc.crypto.mceliece.McElieceKobaraImaiCipher;
    import java.security.SecureRandom;
    import java.util.Arrays;

    public class Encrypter {

        public static final SecureRandom SECURE_RANDOM;

        static {
            SecureRandom tmp = null;
            try {
                tmp = SecureRandom.getInstance("SHA1PRNG");
            } catch (Exception nsae) {
                throw new RuntimeException(nsae);
            } finally {
                SECURE_RANDOM = tmp;
            }
        }

        public static void main(String[] args) throws Exception {
            // Generate random data
            byte[] dataToEncrypt = new byte[250];
            SECURE_RANDOM.nextBytes(dataToEncrypt);

            // Generate key pair
            AsymmetricCipherKeyPair keyPair = generateMcElieceKeyPair(new McElieceCCA2Parameters());

            // Repeat a few times, because the error will not occur on every run
            for (int iteration = 0; iteration < 1000; iteration++) {

                byte[] encrypted = encryptMcEliece(dataToEncrypt, keyPair.getPublic());
                byte[] decrypted = decryptMcEliece(encrypted, keyPair.getPrivate());

[...]

PR #944, opened 2021-04-29
Fix McEliece KobaraImai decryption error when c5 has leading zeroes

tonywasher:
Fixes issue #942. 

When c5 is split out from c2c1 in the encryption phase, it will occasionally have leading zeroes
On decryption when c5 is derived from the cipher Text, leading zeroes are stripped. This results in an incorrect decryption. This patch detects a short c5 and extends it to the correct length by restoring any stripped zeroes.


dghgit:
Thanks for the patch, merged! I have added the CCA2 keys to the PrivateKey/PublicKeyFactory's as well.

Issue #153, opened 2016-08-20
[ASN.1] No support for writing PRIVATE tag


swalendzik:
PRIVATE tags are are mapped into DERApplicationSpecific object so write ASN.1 bytes are differen

example
i have ASN.1 : C00474657374 
`ByteArrayInputStream asnPrivateTag = new ByteArrayInputStream(Hex.decode("C00474657374"));`
`ASN1Primitive asn1Primitive = new ASN1InputStream(asnPrivateTag).readObject();`
`ByteArrayOutputStream outputAsn = new ByteArrayOutputStream();`
`new ASN1OutputStream(outputAsn).writeObject(asn1Primitive);`
After write to output stream i got ASN.1 : 400474657374

avin98:
Whats the workaround for this ?

swalendzik:
You can create DERApplicationSpecific with tag number - 64

example
``
ByteArrayInputStream asnPrivateTag = new ByteArrayInputStream(Hex.decode("C00474657374"));
DERApplicationSpecific asn1Primitive = new ASN1InputStream(asnPrivateTag).readObject();
ByteArrayOutputStream outputAsn = new ByteArrayOutputStream();

DERApplicationSpecific privateTag = new DERApplicationSpecific(asn1Primitive.isConstructed(), asn1Primitive.getApplicationTag() - 64, asn1Primitive.getContents())
new ASN1OutputStream(outputAsn).writeObject(asn1Primitive);
``

amannm:
Yes I was curious why the library was missing this class. The "workaround" seems to work for writing, but it would be better to have an explicit class for private tags coming in.

dghgit:
I've merged 959 with a few corrections. Marking this one as now resolved.

PR #959, opened 2021-05-22
add support for ASN.1 private tags

amannm:
The ASN.1 standard defines a total of 4 tag classes: 
* Universal
* Application Specific
* Context Specific
* Private

This PR adds support for the currently absent "Private" tag class to the bc-java ASN.1 library and addresses issue #153


dghgit:
Thanks for the patch. Merged with a few corrections. Should appear on github shortly.

PR #927, opened 2021-04-13
Fixed AlgorithmIdenifier.parameters of digest algo, extended ECIES, added XIES

xipki:
This is an amendment to the pull #905.

PR #961, opened 2021-05-27
Fixed AlgorithmIdenifier.parameters of digest algo (split pull request #927)

xipki:


dghgit:
I think this one is finally dealt with. The TSP server should really be echoing back the algid requested, or send bad_alg if it doesn't recognise it. I'm not sure why we broke out the OBJECT IDENTIFIER in the imprint request originally. Using the actual algId will be come important if anyone starts using things like shake12len or similar.

PR #962, opened 2021-05-27
extend ECIES with SHA256, SHA384 and SHA512

xipki:
Extend ECIES (split pull request #927)


dghgit:
Merged. Thanks for the patch!

PR #966, opened 2021-05-30
Add implementation of XIES

xipki:
split pull request #927

### DIFF
--- a/prov/src/test/java/org/bouncycastle/jce/provider/test/XIESTest.java
+++ b/prov/src/test/java/org/bouncycastle/jce/provider/test/XIESTest.java
@@ -1,0 +1,307 @@
+package org.bouncycastle.jce.provider.test;
+
+import java.security.InvalidAlgorithmParameterException;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.SecureRandom;
+import java.security.Security;
+
+import javax.crypto.BadPaddingException;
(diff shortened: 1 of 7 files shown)


prov/src/test/java/org/bouncycastle/jce/provider/test/XIESTest.java
```diff
@@ -0,0 +1,307 @@
...
+            {
+                c2.doFinal(tmp, 0, tmp.length);
+                fail("decrypted corrupted data");
+            }
+            catch (BadPaddingException e)
+            {
+                isTrue("wrong message: " + e.getMessage(), "unable to process block".equals(e.getMessage()));
+            }
+        }
+// TODO: DHAES mode is not currently implemented, perhaps it shouldn't be...
```
kenluck2001, line 273: Please remove the comments as it is not implemented as it may be merged. Create a separate issue to track that issue if it is useful to be implemented in the future

xipki: Sorry I deleted the fork of bc-java, and cannot update the code based on this PR. Could you please delete these commented lines before/after the merge?

kenluck2001: I am not a committer, but just reading through the PR.

dghgit:
Thanks for the patch. I've added this with some changes. I created a version of IESCipher specifically for EdEC and I left out the Triple-DES variants. People should really be getting away from T-DES, but if this is likely to cause issues let me know.

PR #970, opened 2021-06-10
PGP{Public|Secret}KeyRingCollection: Ignore marker packets when readi…

vanitasvitae:
…ng from inputStream

According to [RFC4880](https://datatracker.ietf.org/doc/html/rfc4880#section-5.8) marker packets must be ignored when received.

This PR ignores marker packets when encountered while parsing PGPPublicKeyRingCollections and PGPSecretKeyRingCollections from input streams.

Also, see https://tests.sequoia-pgp.org/#Marker_Packet


mwcw:
Thanks Paul,

This has been merged and will be synced up later.

MW

PR #980, opened 2021-06-18
Create PGPSignatureSubpacketVector factory method

vanitasvitae:
This PR adds a factory method to the `PGPSignatureSubpacketVector` class that allows instantiation of an object without the need to use the fiddly `PGPSignatureSubpacketGenerator` class.


kenluck2001:
It seems that that the line `packets = new SignatureSubpacket[0];`, `SignatureSubpacket` does not seems to have scope here as it is not passed in this method and is probably undefined.

vanitasvitae:
What do you mean?
The check makes sure that if the user passes `null` as an argument for `packets`, the constructor is instead called with an empty array.

dghgit:
Thanks for the patch. Merged! Will appear in 170b08

PR #988, opened 2021-07-09
Add SHAKE256-len and SHAKE128-len digest to BcDefaultDigestProvider

matheus-eyng:
Although there is support for shake256-len and shake128-len in other
 classes throughout the project, it's still missing from the BcDefaultDigestProvider. 

This can cause a lot of headache when trying to sign certain types of CMS.

### DIFF
--- a/pkix/src/main/java/org/bouncycastle/operator/bc/BcDefaultDigestProvider.java
+++ b/pkix/src/main/java/org/bouncycastle/operator/bc/BcDefaultDigestProvider.java
@@ -150,6 +150,20 @@ public class BcDefaultDigestProvider
                 return new RIPEMD256Digest();
             }
         });
+        table.put(NISTObjectIdentifiers.id_shake256_len, new BcDigestProvider()
+        {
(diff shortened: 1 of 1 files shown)


kenluck2001:
Please can you refactor the file to use constants instead of the number literals in the source code. `SHAKEDigest(256)` as `SHAKEDigest(CONSTANTVARIABLE)`

matheus-eyng:
I'm a little hesitant to add these constants since I'm not really familiar enough with the whole project. I thought there would be a way to get the length based on the identifier but I couldn't find any and I wanted to make as little changes as possible. Furthermore, I was following the class pattern of calling the constructor with literals since they are part of the algorithm itself.

Some further guidance would be appreciated.

kenluck2001:
Don't worry. I am also new to projects. Leave it as it is

dghgit:
I've added support although I had to make a few changes. The -len OIDs actually  have a parameter - CMS makes it possible to get away with the default as the parameter used is the default, but as a general rule it's not correct to assume that.

Should appear in BC 1.70b11 on https://www.bouncycastle.org/betas in the next day or so.

Issue #998, opened 2021-07-26
Exception in checksum verification in ArmoredInputStream is sometimes swallowed


vanitasvitae:
Hey!

The ArmoredInputStream does a checksum verification in its `read()` method.
Unfortunately, a mismatching CRC checksum results in an `IOException` being thrown.

This is unfortunate, because the default implementation of `InputStream.read(bytes, offset, length)` [silently swallows](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/io/InputStream.java#L286) `IOExceptions`. Therefore the CRC checksum error goes unnoticed.

I see two ways of fixing this:
1. Throw a custom exception not derived from `IOException`. This will likely break existing implementations, as a new runtime exception type needs to be introduced.
2. Override `ArmoredInputStream.read(bytes, offset, length)` to not swallow IOExceptions. This is probably the preferred way to solve the issue.

See https://github.com/pgpainless/pgpainless/issues/159#issuecomment-886694555

dghgit:
Closing as #999 now merged.

PR #999, opened 2021-07-26
ArmoredInputStream: Prevent IOExceptions from being silently swallowed

vanitasvitae:
Java's [default implementation](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/io/InputStream.java#L286) of the InputStream class silently swallows `IOExceptions` when `read(bytes)` or `read(bytes, offset, length)` are used to read from the stream. This causes CRC checksum errors from going unnoticed.

This PR changes `ArmoredInputStream` to override `read(bytes, offset, len)` in order to prevent silently swallowing `IOExceptions`.

Related bug report: https://github.com/bcgit/bc-java/issues/998

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
@@ -488,7 +488,63 @@ public class ArmoredInputStream
 
         return c;
     }
-    
+
+    /**
+     * Reads up to <code>len</code> bytes of data from the input stream into
+     * an array of bytes.  An attempt is made to read as many as
+     * <code>len</code> bytes, but a smaller number may be read.
+     * The number of bytes actually read is returned as an integer.
+     *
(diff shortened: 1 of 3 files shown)


vanitasvitae:
I added a commit that makes the ArmoredInputStream ignore missing CRC checksums by default (i.e. don't throw an IOException), but also allow change of this behavior via `armoredInputStream.setIgnoreMissingCRCSum(false)`.

Motivation of this change is, that https://tests.sequoia-pgp.org/#ASCII_Armor suggests that missing CRC checksums should still pass the parser just fine. Only mismatching CRC sums should present an issue and should be thrown as an IOException.

dghgit:
Thanks for the patch, merged with minor change - I've made setIgnoreMissingCRCSum  setDetectMissingCRC with the default as false (so a missing CRC will not cause an error, in line with what exists as far as guidelines). Will be in 170b08.

PR #1001, opened 2021-07-31
PGPPublicKey: Improve handling of signatures for duplicated user-id/attributes

vanitasvitae:
Since BC internally relies on `List<List<PGPSignature>>` and indices to keep track of which signatures belong to which user-id, it cannot properly handle duplicate user-id packets.

As a consequence, BC will not return all signatures on a duplicated user-id, if the certificate is of the form
```
Public Key
User-id
User-id (duplicate)
User-id Signature
```
since it will return the idSigs item at index 0, which is empty in this case.

This PR changes the behavior of BC by accumulating all signatures that match a given user-id / user-attributes packet before returning all collected signatures. The behavior of returning null if no user-id matches is preserved.


dghgit:
Thanks for the patch. Merged! Will be in 170b08.

PR #1006, opened 2021-08-05
Ignore PGPSignatures with invalid version

vanitasvitae:
This PR modifies the `PGPObjectFactory`'s behavior such that it ignores `PGPSignature` objects with an invalid version number.
That way we can continue processing data which contains part known, part unknown signatures.

This test makes BC behave the way the [sequoia-pgp interop test suite](https://tests.sequoia-pgp.org/#Detached_signatures_with_unknown_packets) expects.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPObjectFactory.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPObjectFactory.java
@@ -95,6 +95,17 @@ public class PGPObjectFactory
                 {
                     l.add(new PGPSignature(in));
                 }
+                catch (RuntimeException e)
+                {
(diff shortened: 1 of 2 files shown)


dghgit:
Thanks for the patch. Merged with revision (added a specific exception). Will appear in 170b08

dkg:
For those following up on this (including anyone looking to backport to fix an older stable version), I think it landed upstream with the following two commits: 

- 76e8043bec1d4b590492fe187c1136aaae6778ed (@vanitasvitae's original contribution)
- 402920e90a3d88c01c1308d472a2d4071761a4c9 (@dghgit's cleanup)

Thanks to both of you for doing this work.

PR #1007, opened 2021-08-08
ArmoredInputStream: Skip '\v\f' in blank lines

vanitasvitae:
When parsing ASCII armored data, it is advisable to skip certain whitespace characters. BC already correctly skips `\r\n` and `\t\r`.
This PR modifies the behaviour of `ArmoredInputStream` to also skip `\t\r\v\f`.

As a result, BC's ASCII armor implementation will agree with the expectations of the [OpenPGP Interoperability Test Suite](https://tests.sequoia-pgp.org/#Mangled_ASCII_Armored_Key).


dghgit:
Thanks for the patch, merged! Will be in 170b08 or later.

PR #1036, opened 2021-10-01
Remove support for maxXofLen in Kangaroo digest

tonywasher:
The Kangaroo digest has been implemented with a maxXofLen feature. Unfortunately the default for this is the natural underlying digest length, rather than unlimited as other BouncyCastle code (noticeably PSSSigner) assumes.

Since the maxXofLen does not affect the output (as in Blake2Xs), it is a rather redundant and unnecessary feature. In addition it is out of alignment with the SHAKE XOF implementation.

This pullRequest removes the maxXofLen parameter and it's policing.


mwcw:
Thanks,

I updated the test as well.
This has been merged and will be synced a little later on.

PR #1037, opened 2021-10-03
Police Blake3 output limit

tonywasher:
The Blake3 Specification sets a limit to the XOF output of 2<sup>64</sup>-1 bytes. The current implementation does not police this limit. This pull request adds logic to police this limit.

This pull request is provided for completeness only. During testing of this change I found that it takes around a second to process 2<sup>28</sup> bytes.  Given there are approximately 2<sup>26</sup> seconds in a year, it will take around 1000 years to hit this limit!!


mwcw:
Thanks,

This has been merged and will be synced later.

MW

PR #1040, opened 2021-10-06
Fix CRLF encoding in LiteralData packets with TEXT or UTF8 encoding

### LINKED ISSUE #1035: PGPLiteralDataGenerator: Text mode does not convert line endings to <CR><LF>
As pointed out by @bjansen [here](https://github.com/pgpainless/pgpainless/discussions/191), Bouncycastle does not convert line endings of data inside a Literal Data packet to <CR><LF>, even though the format is set to `t` (text mode).

I'm not sure, how easy it would be to fix this issue. Maybe it is enough to wrap the `WrappedGeneratorStream` into a `FilteredOutputStream` which replaces the line endings, before returning the stream in `PGPLiteralDataGenerator.open()`?

vanitasvitae:
Fixes #1035

One thing that bothers me is, that [PGPLiteralDataGenerator.open()](https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/pg/src/main/j2me/org/bouncycastle/openpgp/PGPLiteralDataGenerator.java#L92) requires the user to input the length of the canonicalized data.
This is a source for error, as the user probably will input `uncanonicalizedData.length` here, which will result in a broken message (see the fixed tests for examples).

Can we hide this method maybe and rely solely on the method that takes a buffer instead?

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPLiteralData.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPLiteralData.java
@@ -130,4 +130,17 @@ public class PGPLiteralData
         }
         return this.getInputStream();
     }
+
+
+    /**
+     * Return the input stream representing the data stream.
+     * Equivalent to {@link #getInputStream()}.
+     */
+    public InputStream getDataStream(String lineSeparator)
+    {
+        if (getFormat() == TEXT || getFormat() == UTF8) {
+            return new CRLFDecoderStream(this.getInputStream(), lineSeparator);
+        }
+        return this.getInputStream();
+    }
 }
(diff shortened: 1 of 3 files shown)


dghgit:
Tricky.

Yes, length is one issue with changing the data that's sent. If people need to include properly canonicalized data in LiteralData the correct approach would be to provide a utility/wrapper which canonicalizes the data and put it in the literal data. Tampering with the actual data provided to the literal data object, particularly because someone else has failed to implement the standard correctly, would always be a bad idea.

I think this could be accommodated in the API for both a streaming look and straight out file model, as you say, you need the length in advance if you don't wish to use indefinite-length stream, but a wrapper that follows the convention of providing a buffer to signify indefinite-length streams might work. Perhaps a PGPCanonicalizedDataGenerator would be the place to start - this would just need to wrap PGPLiteralDataGenerator and either buffer in-memory/to-a-backing-file for length and write indefinite length packets otherwise. Hmmm. Give me a day to look at this one.

dghgit:
Okay, I've just pushed PGPCanonicalizedDataGenerator which explicitly transforms the data feed into it when creating the literal data object. I think this will provide an easy way of working with broken servers as well as not causing any unpleasant surprises for people using the old API. Let me know what you think.

dghgit:
I've just uploaded one, see https://www.bouncycastle.org/betas 170b09 or later should do the trick.

PR #1042, opened 2021-10-15
Make StreamGenerator interface public

vanitasvitae:
Since both `PGPLiteralDataGenerator` and `PGPCanonicalizedDataGenerator` inherit from this interface, making it public would allow downstream users to use a variable of type `StreamGenerator` to hold either the `PGPLiteralDataGenerator` or `PGPCanonicalizedDataGenerator` in one field.

As such the user can do stuff like:
```
private StreamGenerator streamGenerator;
[...]

encode() {
    if (encoding == BINARY) {
        outputStream = literalDataGenerator.open([...]);
        streamGenerator = literalDataGenerator;
    } else {
        PGPCanonicalizedDataGenerator canonicalizedDataGenerator = new PGPCanonicalizedDataGenerator();
        outputStream = canonicalizedDataGenerator.open([..});
        streamGenerator = canonicalizedDataGenerator;
    }
    [...]
    streamGenerator.close();
}

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/StreamGenerator.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/StreamGenerator.java
@@ -6,7 +6,7 @@ import java.io.IOException;
  * Callback interface for generators that produce a stream to be informed when the stream has been
  * closed by the client.
  */
-interface StreamGenerator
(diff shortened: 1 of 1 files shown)


vanitasvitae:
I found another way to work around this by using a wrapper class.

PR #1047, opened 2021-10-22
Extend BcDefaultDigestProvider for algorithms: SHA512-224, SHAKE512-256, SHAKE128, SHAKE256, SHAKE128Len, SHAKE256Len, and SM3; Resolve the name for SHAKE based signature

xipki:


xipki:
@dghgit Could you also add the support of the hash algorithm SM3?

xipki:
@dghgit In the commit 56bb1e3, you need to change SHAKE512 to SHAKE256.

dghgit:
Thanks for the patch. There was a bit of revising while merging as it overlapped another one, but I think everything in this one is now covered as well. Updates should appear on github shortly and in 170b12.

PR #1055, opened 2021-11-03
Create PGPKeyRingGenerator constructor for key rings without user-id

vanitasvitae:
Hey!

Since an OpenPGP keyring does not necessarily require a user-id, this PR adds a new constructor for the `PGPKeyRingGenerator` which takes a primary key but no user-id.

Instead of creating a user-id certification, this constructor places a direct-key signature on the primary key.


dghgit:
Thanks for the patch! Merged.

PR #1063, opened 2021-11-19
Add factory method to instantiate a public subkey from a public key packet

vanitasvitae:
Unfortunately, the only public constructor for public keys does not set
subSigs to a non-null value. Since isMasterKey() derives its result
from the nullity of the subSigs list, it is not possible to create
public subkey objects without going a detour over the KeyRingGenerator.
This commit adds a factory method which sets the subSigs list to an
empty but non-null list.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKey.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKey.java
@@ -148,6 +148,19 @@ public class PGPPublicKey
         init(fingerPrintCalculator);
      }
 
+    /**
+     * Factory method to create a subkey from the given public key packet.
+     *
+     * @param publicKeyPacket public key packet
+     * @param fingerPrintCalculator fingerprint calculator
+     * @return subkey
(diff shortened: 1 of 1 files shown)


vanitasvitae:
It might also be a good idea to change the logic of `pgpPublicKey.isMasterKey()` to calculate its result based on the packet type instead of what it currently does, but I'm not sure if there is a reason for the current behavior.

vanitasvitae:
I'm not sure if this PR is the right way to solve my issue.
My problem is, that I'm unable to generate a PGP subkey without using the PGPKeyRingGenerator.
I came up with my own logic for binding signature creation which cannot use BCs PGPKeyRingGenerator but I cannot find any public constructor to create a subkey. Hence my PR.

dghgit:
Okay, it took me a while to get my head around this one - I think the real answer might be in your second comment on the 19th. I've fixed the constructor to process the public key as well according to the boolean. I'm not sure if this is 100% of the answer, but it's a lot closer to the answer than where we were before.

PR #1084, opened 2022-01-04
Simplify PGPPublicKey.isMasterKey()

vanitasvitae:
The current implementation of `PGPPublicKey.isMasterKey()` is a bit unstable.
This PR replaces it with a more straight forward implementation that inspects the public key packet to infer the key type.


dghgit:
Thanks for the patch. I left the check for isEncryption/RSA_GENERAL in, the method needs to recognise that a particular key can't be a master key, despite what someone's carefully handcrafted public key packet claims it is. Merged.

PR #1085, opened 2022-01-04
SignatureSubpacketInputStream: return RevocationKey subpacket

vanitasvitae:
When trying to extract `RevocationKey` subpackets from a `PGPSignatureSubpacketVector`, I ran into `ClassCastingException`s.
Turns out the `SignatureSubpacketInputStream` was missing a branch to parse `RevocationKey` subpackets.


dghgit:
Thanks for the patch. Merged!

PR #1088, opened 2022-01-12
Add constant for MIME encoding to PGPLiteralData class

vanitasvitae:
This format is introduced in rfc4880-bis10

See https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-10.html#name-literal-data-packet-tag-11


dghgit:
Thanks for the patch! Merged. I've got a feeling this is going to be interesting.... text should have been enough, I wonder what they've got in mind, I guess we'll find out soon enough.

dghgit:
This change, as well as the earlier ones is now up in https://www.bouncycastle.org/betas 171 b04 or later.

PR #1089, opened 2022-01-12
Algorithm parameters lost on fallback

schaeferm:
I have noticed that the algorithm parameters get lost when using the fallback to encrypt/decrypt mode. This doesn't work for example with RSA OAEP when non default parameters are used.


dghgit:
Thanks for the patch, nice pickup! Merged.

dghgit:
This is now up in https://www.bouncycastle.org/betas 171 b04 or later.

PR #1090, opened 2022-01-17
Add support for Regex subpackets

vanitasvitae:


core/src/main/java/org/bouncycastle/util/Strings.java
```diff
@@ -56,6 +56,26 @@ public static String fromUTF8ByteArray(byte[] bytes)
         return new String(chars, 0, len);
     }
 
+    public static String fromNullTerminatedUTF8ByteArray(byte[] bytes)
+    {
+        // Tolerate strings which are non-null terminated
```
vanitasvitae, line 61: The RFC [states](https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.14) that the regex is a null-terminated string.
I'm not sure, how BC should handle parsing strings that aren't null-terminated.

For now I opted to tolerate such strings, but if you think we should instead reject strings without null-terminator, we can do that as well.

dghgit: I think we'll need to reject them - I know this will sound excessively paranoid, but these will probably appear in signatures so we want them to be as canonical as we can. There's already been a bit of fun had by exploiting alternate encodings in ASN.1, while I can't think of a direct exploit for something like this, I'm confident it's only lack of imagination on my part and that there'll be someone out there who's more than up to the job of taking advantage.

vanitasvitae: Alright, I removed the tolerance. Instead, now an IllegalArgumentException is thrown for non-null terminated data, similar to how non-utf8 data would cause an exception.

vanitasvitae: One more question: Right now the method `fromNullTerminatedUTF8ByteArray()` is only called from `RegularExpression.getRegex()`.
So the exception is only raised when the user wants to get the regex string from the subpacket, not when the subpacket is created.
Additionally the user could work around the exception by calling `getRawRegex()`. Is that okay for you, or should I change the code such that null-termination is enforced in the subpacket constructor. If so, is it safe to throw an `IllegalArgumentException` in the RegularExpression subpackets constructor?

dghgit: I'm a fan of enforcing on construction in this case.

dghgit:
Merged with null termination check added. Now up on https://www.bouncycastle.org/betas 171b05 or later

vanitasvitae:
Awesome!

PR #1093, opened 2022-01-22
Handle full PGP partial packets with trailing zero-length packet. Fixes #1027

### LINKED ISSUE #1027: Corrupt PGP stream due to compression, partial body lengths, & 0-length body length
I've been attempting to diagnose a parsing failure of a PGP message with the below structure.

## Summary

I think there may be a bug with how Bouncy Castle wires together the JDK's Inflater class, compression packets that use partial body lengths and end with a non-partial but 0 body length header, and maintaining stream alignment with all of the overlayed BCPGInputStream/PartialInputStream instances it uses.

## Packet Structure

```
  public_key_encrypted_session (RSA_GENERAL)
  public_key_encrypted_session (RSA_GENERAL)
  symmetrically_encrypted_integrity_protected_data (AES256)
      compressed_data (ZLIB, uses partial body lengths)
          one_pass_signature (SHA512/RSA_GENERAL)
          one_pass_signature (SHA512/RSA_GENERAL)
          literal_data (~1300 bytes)
          signature
          signature
      reserved_packet     <--- I think this is an artifact of a misaligned stream
      modification_detection_code (verifies correctly)
```
## Detailed Description

Even after you fully read the stream returned from `PGPCompressedData.getDataStream()`, the overall stream is positioned at that last 0 body length header (which is [encoded](https://datatracker.ietf.org/doc/html/rfc4880#section-4.2.2.1) as a single 0 byte).  When you go to read the next packet tag, it reads the 0, thinks it's the [RESERVED](https://github.com/bcgit/bc-java/blob/master/pg/src/main/j2me/org/bouncycastle/bcpg/BCPGInputStream.java#L242) packet tag, and then [throws](https://github.com/bcgit/bc-java/blob/master/pg/src/main/j2me/org/bouncycastle/bcpg/BCPGInputStream.java#L176) when it realizes the header is corrupt (since all headers have at least [bit 7 set to 1](https://datatracker.ietf.org/doc/html/rfc4880#section-4.2) -- `00000000b` isn't a valid packet tag/header).

If you intentionally ignore the exception, the stream becomes aligned again and you can continue processing the [ModificationDetectionCode packet](https://datatracker.ietf.org/doc/html/rfc4880#section-5.14). That packet even verifies which strongly implies the 0 is *supposed* to be there (or else the hash it computes wouldn't verify correctly).  So the problem isn't that a stray 0 got inserted after the compression packet, it's that the Bouncy Castle streams left off the last byte (which happened to be 0) when processing it.

## Client Code

The code processing this message is a recursive loop over each packet (in the same vein as `gpg --list-packets`). It does this with calls to:

- `BCPGInputStream.nextPacketTag()`
- `BCPGInputStream.readPacket()`
- Packet specific logic like decryption/decompression
- Reading past data by consuming `InputStreamPacket.getInputStream()`

I suspect the minimal repro case is just a single CompressionDataPacket that uses ZLIB, partial body lengths, and ends with a non-partial but 0 length body length.  Feed that to PGPCompressedData and consume the data stream.  You'll see that the underlying stream didn't advance to EOF like it should have (it's one before the end).

I haven't yet been able to craft a repro case besides this one file that I can't share.

## Debugging Attempts

After many attempts at stepping through the code with a debugger, I noticed that [PartialInputStream.loadDataLength()](https://github.com/bcgit/bc-java/blob/master/pg/src/main/j2me/org/bouncycastle/bcpg/BCPGInputStream.java#L359) only ever read partial body lengths, it never took the branch for non-partial body lengths at all.  The [RFC states](https://datatracker.ietf.org/doc/html/rfc4880#section-4.2.2.4):

> The last length header
>  in the packet MUST NOT be a Partial Body Length header.  Partial Body
>  Length headers may only be used for the non-final parts of the
>  packet.
>
>  Note also that the last Body Length header can be a zero-length
>  header.

## Root Cause?

I think this all happens because [Bouncy Castle feeds](https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/pg/src/main/java/org/bouncycastle/openpgp/PGPCompressedData.java#L109) the wrapped PartialInputStream containing the compressed data to the JDK's InflaterInputStream/Inflater class. At least with zlib, the stream appears to be self-delimiting.  There's a `finished` flag in those classes that can get set to true without actually reading EOF from the underlying stream.  That gets propagated to the caller after a `fill()` call that triggers `inflate()` to know the stream is done. Nothing reads from the underlying stream to trigger it to read that last non-partial body length byte.

## Solution?

It seems like PartialInputStream needs to remember that the last body length header it read was a partial one. Something will then need to trigger some sort of flush operation that will trigger that last read before the stream is abandoned and left to go out of scope. It may be as easy as wrapping InflaterInputStream in yet another stream that performs the flush operation after EOF is returned (and before passing the EOF along to the caller).  Do note that you'll have to pass this through the stream PartialInputStream is itself wrapped in (just a BCPGInputStream).

## Why is this Hard to Detect?

For PGP messages that use non-integrity protected encryption, callers get their data and are done.  It doesn't matter that the lowest level stream was one byte away from reaching EOF -- the data stream returned EOF and that's all that matters.  There's almost never another packet after CompressionPackets.

For PGP messages that *do* use integrity protected encryption, there's that ModificationDetectionPacket after the CompressionPacket. However, there's a built-in recovery mechanism when you call `PGPEncryptedData.verify()`... It [automatically reads](https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/pg/src/main/java/org/bouncycastle/openpgp/PGPEncryptedData.java#L141) to the end of the stream for you (skipping over that single 0 byte).

tonywasher:


dghgit:
Thanks for the patch. I had to read the rationale a couple of times, but I think it's correct. Nice work! Merged, and now up in https://www.bouncycastle.org/betas 171b06 or later.

PR #1102, opened 2022-02-02
Add method to join two copies of a public key

vanitasvitae:
This method comes in handy, whenever a public key needs to be merged
with another copy of the same key, e.g. when consuming updates from
a key server.

This implementation does not do any verification on signatures and
blindly merges the data together.
Verification is left to the implementation, just like in other places
of BCs PGP implementation.

I added two flags which modify the methods behavior:
* `joinTrustPackets` controls whether changed trust packets from `copy` will be adopted in the resulting merged key
* `allowSubkeySigsOnNonSubkey` controls whether in case of a non-subkey `key`, subkey signatures on `copy` should be accepted into the resulting merged key, which would transform it into a subkey, which might cause issues with BCs implementation (e.g. primary keys must not have subkey signatures currently, although the standard would not forbid it).

Let me know what you think :)

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKey.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPPublicKey.java
@@ -1133,4 +1133,177 @@ public class PGPPublicKey
 
         return returnKey;
     }
+
+    /**
+     * Merge this the given local public key with another, potentially fresher copy.
+     * The resulting {@link PGPPublicKey} contains the sum of both keys user-ids and signatures.
+     *
+     * If joinTrustPackets is set to true and the copy carries a trust packet,
+     * the joined key will copy the trust-packet from the copy.
+     * Otherwise, it will carry the trust packet of the local key.
+     *
+     * @param key local public key
+     * @param copy copy of the public key (e.g. from a key server)
+     * @param joinTrustPackets if true, trust packets from the copy are copied over into the resulting key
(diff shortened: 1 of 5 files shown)


vanitasvitae:
I would especially be interested in whether there is a better way to check for signature equality than `Arrays.areEqual(sig1.encode(), sig2.encode())`.

vanitasvitae:
I changed signature comparison to comparing the actual cryptographic signature encoding. This should deal with different encodings of the PGPSignature packet I believe.

Furthermore I added a method to merge the unhashed subpacket areas of signatures.

dghgit:
It sounds a little complicated and a little risky... I'd probably be verifying prior to merging. Would I be right if I was to think that this is really about not having things duplicated?

vanitasvitae:
I added some test vectors which were kindly provided by Sequoia-PGP.
The code is passing those tests happily :)

vanitasvitae:
> Would I be right if I was to think that this is really about not having things duplicated?

Yes. User-IDs, Keys and Signatures from both copies of the certificate are copied together into the resulting key.

vanitasvitae:
Is there anything I can do to help this PR get merged? :)

dghgit:
I've started looking at it. Sorry things have been a bit crazy of late.

dghgit:
Merged with a minor change. I dropped the KeyFingerPrintCalculator parameter. Should appear on github shortly.

dghgit:
Now up at https://www.bouncycastle.org/betas 171b18 or later.

vanitasvitae:
Awesome, thanks for taking care of this :)

PR #1115, opened 2022-02-19
Add support for PolicyURI signature subpacket

vanitasvitae:
This PR adds support for the [Policy URI Signature Subpacket](https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.20).

TODO:
* [x] Write tests


dghgit:
Thanks for the patch. Merged!

Issue #1120, opened 2022-02-27
Umlauts in ArmoredOutputStream headers are not properly encoded


vanitasvitae:
When I add a header such as a comment to the `ArmoredOutputStream` which contains umlauts, the resulting ASCII armored output did not properly encode the header value.

The following test demonstrates this behavior:
```
package org.bouncycastle.openpgp.test;

import org.bouncycastle.bcpg.ArmoredOutputStream;
import org.bouncycastle.util.test.SimpleTest;

import java.io.ByteArrayOutputStream;
import java.nio.charset.Charset;

public class ArmoredOutputStreamUTF8Test extends SimpleTest {

    private static final Charset utf8 = Charset.forName("UTF8");

    public static void main(String[] args) {
        runTest(new ArmoredOutputStreamUTF8Test());
    }

    @Override
    public String getName() {
        return ArmoredOutputStreamUTF8Test.class.getSimpleName();
    }

    @Override
    public void performTest() throws Exception {
        String utf8WithUmlauts = "ÄÖÜß&Umlauts";

        ByteArrayOutputStream out = new ByteArrayOutputStream();
        ArmoredOutputStream armorOut = new ArmoredOutputStream(out);
        armorOut.setHeader("Comment", utf8WithUmlauts);

        armorOut.write("Foo\nBar".getBytes(utf8));

        armorOut.close();

        String armoredOutput = out.toString();
        String[] lines = armoredOutput.split("\n");
        String comment = null;
        for (String line : lines) {
            if (line.startsWith("Comment: ")) {
                comment = line.substring("Comment: ".length());
                break;
            }
        }

        isTrue("We did not find the comment line. This MUST never happen.", comment != null);
        isEquals("Comment was not properly encoded. Expected: "  +utf8WithUmlauts + ", Actual: " + comment, comment, utf8WithUmlauts);
    }
}
```

The test results in the following error:
```
ArmoredOutputStreamUTF8Test: Comment was not properly encoded. Expected: ÄÖÜß&Umlauts, Actual: ����&Umlauts
```

I tracked down the issue of this bug to the [ArmoredOutputStream using](https://github.com/bcgit/bc-java/blob/master/pg/src/main/java/org/bouncycastle/bcpg/ArmoredOutputStream.java#L308-L311) `out.write(value.charAt(i));` when writing header values, which causes encoding issues.

Compare:
```
        String utf8 = "ÄÖÜß";

        ByteArrayOutputStream out = new ByteArrayOutputStream();
[...]

PR #1121, opened 2022-02-27
Add test for UTF8 encoding issues in ASCII armor

vanitasvitae:
This PR contains the test from #1120


vanitasvitae:
I added a test verifying that comment headers do not survive an ASCII armor round-trip.

dghgit:
Thanks for the test. Now added.

PR #1132, opened 2022-03-23
Implement Argon2 S2K mechanism

vanitasvitae:
This PR aims to implement the new Argon2 S2K mechanism for OpenPGP.
The PR further contains some updates from the crypto-refresh revision of the OpenPGP spec.

~This is still pretty much work in progress, so do not yet merge!~

I added factory methods for different S2K methods to the S2K class.
This should make creating S2K instances more straight forward, since the S2K method is no longer derived from argument ordering.

This also enables creating S2Ks using GNU_DUMMY_S2K methods.


vanitasvitae:
Ready for review :) I can squash the commits if necessary.

vanitasvitae:
Squashed where sensible.

dghgit:
Thanks for the patch! Merged.

PR #1166, opened 2022-05-18
Fix parsing of certs, keys and messages with unrecognizable subkeys, signatures and ESKs

vanitasvitae:
BC currently chokes when parsing OpenPGP certificates with contain subkeys with unrecognized algorithms.
As a result, the whole PGPPublicKeyRing becomes null and is therefore unusable.

This means that unfortunately BCs OpenPGP implementation is not upwards compatible to when keys with PQC subkeys are introduced. Ideally an implementation should just ignore subkeys that it does not recognize and instead try to use the other keys.

This PR changes BCs behavior to ignore unrecognizable subkeys when parsing PGPPublicKeyRing and PGPSecretKeyRing objects.
Note, that only subkeys are ignored, since a key with an unrecognizable primary key is not usable at all.

I took the test vectors from [the OpenPGP interoperability test suite](https://tests.sequoia-pgp.org/#Mock_PQ_subkey) and implemented them as unit tests.
Unfortunately there weren't any test vectors for secret keys available, but the change is similar enough as to that I have confidence in its correctness.

In any case I made an extra commit for the changes in secret key parsing.

I also added commits that ignore unsupported signatures when reading then from certs and from ObjectFactory objects.
Further, the PGPObjectFactory will no longer fail to parse PKESKs/SKESKs when there are unknown versions in the list.

This is further improving upwards compatibility.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyEncSessionPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyEncSessionPacket.java
@@ -22,6 +22,10 @@ public class PublicKeyEncSessionPacket
         throws IOException
     {      
         version = in.read();
+        if (version != 3)
+        {
+            throw new UnsupportedPacketVersionException("Unsupported PGP public key encrypted session key packet version encountered: " + version);
+        }
         
         keyID |= (long)in.read() << 56;
         keyID |= (long)in.read() << 48;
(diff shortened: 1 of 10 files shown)


dghgit:
Yes, looking at the new draft and what's appearing because of it I can see the merit in this... I've added the option of logging where things are being skipped so people can keep an eye on what might be going wrong if they need. Merged, thanks!

Issue #1176, opened 2022-06-10
Wrong OIDs in JavaDoc for NISTObjectIdentifiers


mgrafl:
In `NISTObjectIdentifiers` the constants for RSA PKCS \#1 v1.5 Signature with SHA-3 family have different OIDs in their JavaDoc comments than what the actually specify, (e.g., `id_rsassa_pkcs1_v1_5_with_sha3_224` is specified as OID "2.16.840.1.101.3.4.3.13" but is documented as "2.16.840.1.101.3.4.3.9".

https://github.com/bcgit/bc-java/blob/7f4a0a69c19c516c9c154776128e284e6816ecf8/core/src/main/java/org/bouncycastle/asn1/nist/NISTObjectIdentifiers.java#L220-L228

PR #1177, opened 2022-06-10
Correct OIDs in JavaDoc

mgrafl:
Fix for #1176.


dghgit:
Thanks for the patch. Merged!

PR #1179, opened 2022-06-22
Implement buffered read() methods for TruncatedStream

vanitasvitae:
Hey,
This PR adds implementations of `read(byte[] buf)` and `read(byte[] buf, int off, int len)` to `PGPEncryptedData.TruncatedStream`.

This class did not have those methods implemented before, which resulted in degraded performance when decrypting.
The added methods improve performance of the stream.

Previously decrypting a 1GB file took 27 seconds, while the changes in this PR result in an improved time of ~20 seconds on my machine.

Currently the new read() methods read [up to 8192 bytes](https://github.com/bcgit/bc-java/pull/1179/files#diff-83881d5485ba95e07b97362e957d08ea04fe479d7ed772db5c8db5336d1545d8R88) per call. Let me know if I should change the implementation such that they try to exhaust whatever `len` value is given.


dghgit:
Thanks for the patch, merged!

PR #1183, opened 2022-06-29
Do not encode PGPSignatures marked with exportable=false if forTransf…

vanitasvitae:
…er is true

This PR aims to make BC's OpenPGP implementation compatible with [section 5.2.3.11](https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.11) of RFC4880:

```
   Non-exportable, or "local", certifications are signatures made by a
   user to mark a key as valid within that user's implementation only.

   Thus, when an implementation prepares a user's copy of a key for
   transport to another user (this is the process of "exporting" the
   key), any local certification signatures are deleted from the key.
```

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
@@ -599,6 +599,12 @@ public class PGPSignature
         boolean forTransfer)
         throws IOException
     {
+        // Exportable signatures MUST NOT be exported if forTransfer==true
+        if (forTransfer && (!getHashedSubPackets().isExportable() || !getUnhashedSubPackets().isExportable()))
+        {
+            return;
+        }
+
         BCPGOutputStream out;
 
         if (outStream instanceof BCPGOutputStream)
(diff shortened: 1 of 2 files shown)


pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
```diff
@@ -599,6 +599,12 @@ public void encode(
         boolean forTransfer)
         throws IOException
     {
+        // Exportable signatures MUST NOT be exported if forTransfer==true
+        if (forTransfer && (!getHashedSubPackets().isExportable() || !getUnhashedSubPackets().isExportable()))
```
vanitasvitae, line 603: I am up for debate, whether or not we should respect Exportable subpackets in the unhashed section of the signature.

dghgit: I think, until we have information suggesting we shouldn't, we should. It's easy to add an exception later, it can be very hard to withdraw an exception once it's made though.

dghgit:

pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
dghgit: Yes, agreed. I just thought I'd mention it, in theory unhashed packets can be added and removed at will, but it wouldn't be the first time theory does not quite match with reality. I'd say what's just been added is an improvement on before at least, that's the main thing.

vanitasvitae: I talked to some people at Sequoia-PGP and they think that simply wiping the unhashed area is not a good idea.
Instead they proposed to add a `normalize()` method, which removes subpackets following some well-defined rules:

PR #1188, opened 2022-07-14
Improve type safety and exception throwing of PGP*KeyRingCollection

vanitasvitae:
I added generics to internal data structures of `PGPPublicKeyRingCollection` and `PGPSecretKeyRingCollection` classes and also removed `throws PGPException` from many methods.

It turns out, while this exception was declared, it is actually not thrown in any of these methods, so removing it makes using these classes downstream easier.


dghgit:
It's about time, thanks for the patch, merged with minor revision.

dghgit:
Your assumption is correct. Thanks, it would things easier.

PR #1189, opened 2022-07-14
Fix possible bug in PGPPublicKey.removeCert()

vanitasvitae:
The method looped over a list and removed from it.
This can lead to items being skipped.
I fixed it by inverting the loop direction.

This is probably not a big deal in real-world usage, since duplicates in `returnKey.ids` are unlikely, still its probably better to fix it nonetheless.


dghgit:
Yes, this will definitely not do any harm. Thanks, merged.

PR #1190, opened 2022-07-15
Generify ArmoredOutputStream, add support for multiple + new hash algorithms

vanitasvitae:
This PR adds generics to data structures used inside the ArmoredOutputStream.

Further, for the Cleartext Signature Framework it enables having multiple "Hash: " header entries. This change is backwards-compatible by using the varargs notation for hash headers.

Lastly, it adds support for hash headers of new algorithms of the SHA3 family.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredOutputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredOutputStream.java
@@ -6,6 +6,7 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Iterator;
+import java.util.List;
 
 import org.bouncycastle.util.Strings;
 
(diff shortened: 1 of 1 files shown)


dghgit:
Merged with minor changes. Thanks!

vanitasvitae:
Unfortunately, this change introduced some regressions in Debian:
https://tracker.debian.org/pkg/bouncycastle
E.g. https://ci.debian.net/data/autopkgtest/testing/amd64/p/pgpainless/29251802/log.gz

While I'm not sure if there needs something to be done about this, I thought it was worth pointing this out for sake of completeness.

My guess as to why the regression happens is is as follows:
When I changed the method signature from `ArmoredOutputStream.beginClearText(int)` to `ArmoredOutputStream.beginClearText(int...)`, this introduced a "hidden" ABI breakage between 1.71 and 1.72, since `beginClearText(int)` can no longer be found.
To me this error was hidden away, since bumping the BC dependency locally and recompiling my project fixed the issue transparently (no change in my code was necessary to switch from the int-method to the varargs method).

However, this means that consumers compiled against BC 1.71 can not be used with BC 1.72 without recompiling.
Am I right?

dghgit:
Yes, so it thinks int[], annoying, I'll add a backwards compatible call for 1.73, although really I think people are better off if they recompile in general.

PR #1191, opened 2022-07-15
Fix PGPSessionKey.toString()

vanitasvitae:
Calling .toString() on the raw array `sessionKey` produces output which is not helpful.

I'm pretty sure in my original implementation I had `toString()` call `Hex.toHexString()` in place. Was this changed for a reason?

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSessionKey.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSessionKey.java
@@ -30,7 +30,7 @@ public class PGPSessionKey
 
     public String toString()
     {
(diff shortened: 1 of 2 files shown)


dghgit:
Yes, it was in the original patch. The problem with toString() in this context is that it's very easy for something to call it by accident so in general we avoid feeding out either private key data or secret key data via it. "sessionKey" by itself does provide enough information to tell that a particular session key is used, which allows for a level of debugging and tracing, it just doesn't tell you what the session key value is, which can allow for a lot more than just debugging...

vanitasvitae:
Ah I see, that makes sense.

PR #1192, opened 2022-07-15
OpenPGP Code Cleanup: Remove redundant modifiers in interface members

vanitasvitae:
This PR removes redundant modifiers on members of interfaces.

In Java, method signatures in interfaces are `public` and `abstract` by default.
Constants in interfaces are `public`, `static` and `final` by default.


dghgit:
A bit late for a spring clean! Being um-ing and ah-ing about doing this for a while. Thanks, merged.

PR #1195, opened 2022-07-20
Add fast CRC-24 implementation

vanitasvitae:
The default CRC-24 implementation follows RFC4880, which proposes the use of
an iterative algorithm. Mimicking a feedback shift register in software is inefficient.

In 'Computation of Cyclic Redundancy Checks via Table Look-Up'
Dilip V. Sarwate proposes the use of lookup tables to speed up computation.

This PR implements such table lookup.
By calling CRC24.setUseFastImplementation(true), the user can choose to
use the fast implementation instead of the traditional one.

The fast algorithm can process 200mb of data in 3.6 seconds, compared to the iterative one which takes 4.9 seconds for the same data.


dghgit:
Merged with revision. To be honest I did actually like the interface, I just worried about changing the API for this one - I've added FastCRC24 instead. It's now used by default. I can report much better figures though - it appears to be about ten times faster. Thanks!

PR #1219, opened 2022-08-31
Add missing SHA3 digests to BcImplProvider.createDigest()

vanitasvitae:
Hey,
I noticed that implementations for the SHA3 algorithm family were missing from the `createDigest()` method, so I added them here.
I hope I used the right Digest class and instantiation method.


dghgit:
Thanks. Merged!

PR #1227, opened 2022-09-10
Expose PGPObjectFactory.nextObject(BCPGInputStream, KeyFingerprintCal…

vanitasvitae:
…culator)

By exposing the nextObject() method as a static method, it can be used more flexibly.

The reason for this PR is, that `PGPObjectFactory` wraps any stream it gets as constructor argument into a `BCPGInputStream`, regardless of whether that stream might already be a `BCPGInputStream`.
That has the effect, that the stream that got passed into the factory is forwarded too far when parsing objects.

By exposing the `nextObject()` method statically, we have the option to avoid the wrapping, so that the stream is only read so far as needed.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPObjectFactory.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPObjectFactory.java
@@ -81,6 +81,12 @@ public class PGPObjectFactory
     public Object nextObject()
         throws IOException
     {
+        return nextObject(this.in, this.fingerPrintCalculator);
+    }
+
+    public static Object nextObject(BCPGInputStream in, KeyFingerPrintCalculator fingerPrintCalculator)
+            throws IOException
+    {
         List l;
 
         switch (in.nextPacketTag())


vanitasvitae:
Maybe some background to why I created this PR:
I'm working on an implementation that enforces that OpenPGP messages follow the grammar defined in the [RFC](https://www.rfc-editor.org/rfc/rfc4880#section-11.3).

I'm doing this by manually iterating over the packets in a `BCPGInputStream`, updating a pushdown automaton to verify the message syntax. Now, when I want to parse out a PGP* Object (as opposed to a *Packet), the only way to do so currently is to use the PGPObjectFactory class (since most PGP* constructors are not accessible from outside of BC).
Creating a PGPObjectFactory from the BCPGInputStream however results in the issue described above, therefore I'd appreciate the proposed change :)

An alternative solution would be to check if the InputStream argument in the PGPObjectFactory constructor is already a BCPGInputStream. If it is, we can set the inputstream field in the PGPObjectFactory to the constructor argument directly instead of wrapping it.
What do you think?

peterdettman:
> An alternative solution would be to check if the InputStream argument in the PGPObjectFactory constructor is already a BCPGInputStream. If it is, we can set the inputstream field in the PGPObjectFactory to the constructor argument directly instead of wrapping it.

This would be a better choice, as there is already a method `PGPKeyRing.wrap` used for the same purpose by `PGPPublicKeyRing` and `PGPSecretKeyRing` for the same purpose. Also, the bc-csharp code already uses `Wrap` in `PgpObjectFactory` as described (although the method has been moved to `BcpgInputStream`, which might be a good idea here too).

vanitasvitae:
Okay, I'll create another PR for this :)

vanitasvitae:
Closed in favor of #1229

PR #1229, opened 2022-09-12
PGPObjectFactory: Only wrap InputStream with BCPGInputStream if necessary

vanitasvitae:
This PR changes the constructor of the `PGPObjectFactory` class to only wrap the `InputStream` argument inside a `BCPGInputStream`, if the input stream is not a `BCPGInputStream`.

Passing in a `BCPGInputStream` will cause the `in` member of the `PGPObjectFactory` to be assigned to the argument directly.

Alternative solution to the issue described in #1227


dghgit:
Thanks for the patch, merged.

PR #1232, opened 2022-09-21
Expose PGPOnePassSignature.isEncapsulating()

vanitasvitae:
This method exposes the nested flag of OnePassSignatures.
Since the naming was a bit confusing, I renamed the flag to isEncapsulating.
(The rfc names the flag nested, while BC currently interprets `isNested` as "being nested inside", which is the opposite of what the flag is supposed to tell).

Exposing this flag to consumers is required for consumers to verify the signature properly. An encapsulating signature needs to be updated with the encoding of subsequent, nested OPSs and their signatures.

Before this change, consumers would need to encode the signature and access the correct byte to check the flag.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/OnePassSignaturePacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/OnePassSignaturePacket.java
@@ -14,7 +14,7 @@ public class OnePassSignaturePacket
     private int  hashAlgorithm;
     private int  keyAlgorithm;
     private long keyID;
-    private int  nested;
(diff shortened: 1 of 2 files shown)


dghgit:
Thanks for the patch, merged with minor revisions. Encapsulated is also a sadly overloaded term... after much angst I've settled on isContaining(), The problem I had with encapsulated is the data is described as encapsulated which doesn't help clarity either.

vanitasvitae:
As food for thought, the Sequoia-PGP library names this flag `last`. Not sure if that name is better or worse though.
`isContaining` is okay for me too :)

PR #1250, opened 2022-10-08
BCPGInputStream.readPacket(): Consistently call read()

### LINKED ISSUE #1249: BCPGInputStream readPacket() inconsistently calls read()
Hey!
While trying to implement verification of nested OnePassSignatures, I stumbled across a peculiarity of the `BCPGInputStream`.

For my use-case I need to tee out sections of the underlying byte stream to update signatures with it.
My initial idea was to inherit from the `BCPGInputStream`, overriding the `read()` method to tee the byte out.
That way, I could call `readPacket()` and have the packet bytes being automatically written to my signature verifier.
Once a certain packet is reached, I could disable signature updating granularly.

Initially I thought this worked, but for some reason the signatures would not validate properly.
It took a while to find the cause of the issue: For packets with 2byte length encodings, the second length byte would not be written out to my signature verifier.
The reason is that `BCPGInputStream.readPacket()` [internally calls a mix](https://github.com/bcgit/bc-java/blob/master/pg/src/main/java/org/bouncycastle/bcpg/BCPGInputStream.java#L213-L219) of `read()` and `in.read()`, the latter bypassing my overwritten method.

I wonder if it would be worthwhile changing this behaviour in BC. I know this is perhaps a very exotic use-case, and I managed to work around it in PGPainless by using a custom delaying TeeInputStream to wrap the BCPGInputStream around.

vanitasvitae:
Fixes #1249

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/BCPGInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/BCPGInputStream.java
@@ -210,13 +210,13 @@ public class BCPGInputStream
             }
             else if (l <= 223)
             {
-                int b = in.read();
+                int b = this.read();
 
                 bodyLen = ((l - 192) << 8) + (b) + 192;
             }
             else if (l == 255)
             {
-                bodyLen = (in.read() << 24) | (in.read() << 16) |  (in.read() << 8)  | in.read();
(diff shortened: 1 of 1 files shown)


dghgit:
Thanks for the patch. Now merged,

PR #1256, opened 2022-10-16
Add javadoc for PGPPBEEncryptedData.getAlgorithm() and getSymmetricAl…

vanitasvitae:
…gorithm()

From the previous javadoc, it was not totally clear, for which purpose the algorithm would be used.
I tried to improve that with this PR.

Should we maybe rename `getAlgorithm()` to something along the lines of `getKeyEncapsulationAlgorithm()` (although KEMs are usually using asymmetric algorithms)?

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPPBEEncryptedData.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPPBEEncryptedData.java
@@ -42,6 +42,12 @@ public class PGPPBEEncryptedData
         return keyData.getVersion();
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


dghgit:
Thanks for the update. Merged. With the name, it's been there for a while now, so we'd need to duplicate it and add a new version. I'm inclined to leave it alone, the current name is not incorrect, although in hindsight it could be better (that said, even better could be potentially confusing given some of the things heading our way - maybe getKeyWrappingAlgorithm(), but even that's not always true for PBE where in some cases the algorithm is used to generate a key that encrypts the data directly, dammed if you do, dammed if you don't).

PR #1259, opened 2022-10-18
docs: add link to Extension

elexx:
Add a link to Extension because the known ASN1ObjectIdentifiers are located there, hence Extension and Extensions replace X509Extensions.


dghgit:
Thanks for the patch, yes. anything to reduce confusion helps. Merged.

PR #1265, opened 2022-10-21
Ant test output can now show incremental progress

jwcranford:
This is a simple pull request intended to make it a little easier to run the unit tests from the build scripts.

# Problem
No incremental progress is visible at the command-line when running the unit tests via the build scripts. Without monitoring the contents of the test reports directory, it's impossible to tell if the unit tests are hung or just taking a long time.

# Proposed Solution
This pull request provides the ability for a user to enable test summaries from junit; for each test suite executed, ant will print a two-line summary to the console.

# Details
Added ability to override properties in bc-build.properties with a bc-build.user.properties file, which is ignored by git. The intent is for bc-build.user.properties to hold user-specific settings that are *not* committed to the repository.

Test summaries at the console can be enabled by adding this property to bc-build.user.properties:
```
junit.printsummary: on
```

The default behavior remains unchanged, as the default value for junit.printsummary is false.


peterdettman:
Just a minor change for the sake of discoverability: could you please add a brief comment at the top of build.properties pointing to the possibility of build.user.properties overrides.

jwcranford:
Good point - I added a comment to the top of bc-build.properties and merged in the latest from master to resolve a conflict.

dghgit:
Thanks for the suggestion! Now merged.

PR #1266, opened 2022-10-21
Exclude tests

jwcranford:
This is a simple pull request intended to make it easier to run the unit tests locally.

# Problem
Some unit tests take a really, really, really long time to run. It would be nice to be able to temporarily skip them during development.

# Proposed Solution
Both Ant and Gradle build scripts can now ignore unit tests that match a given pattern. Both examples below run all the unit tests *except* those in the pqc package.

## Ant
```
JUNIT_EXCLUDE_TESTS='**/pqc/**' sh build1-8+ test
```

## Gradle
```
gradle test -PexcludeTests="*pqc*"
```


dghgit:
Thanks for the patch. Merged.

PR #1271, opened 2022-10-28
Fix off-by-one error IESCipher#engineGetOutputSize

brandonweeks:
An uncompressed secp192r1 key should be 65 bytes, however engineGetOutputSize() currently returns 64 bytes.


dghgit:
Thanks for the patch. Merged!

PR #1280, opened 2022-11-20
Use Java's rotateLeft in salsaCore to increase likelyhood of inlining

alexbakker:
We use BouncyCastle's scrypt implementation in [Aegis](https://github.com/beemdevelopment/Aegis) (an Android app) to derive a key from the user's password. We recently received [a report](https://github.com/beemdevelopment/Aegis/issues/1024) from a user that key derivation was taking very long on their device (on the order of 2 minutes). The parameters we're using are: ``N=2^15``, ``r=8``, ``p=1``.

Personally, I'm only able to reproduce this when running the app with a debugger attached on Android 13. In my case it takes 30 seconds. Since that user was seeing this on a release build of the app, we figured this was worth investigating. After some profiling, we noticed that a lot of time was being spent calling the ``Integers.rotateLeft`` function in ``Salsa20Engine.salsaCore``. On a hunch I pulled a copy of that class and the ``SCrypt`` class into our codebase and replaced the calls of that wrapper function by directly calling Java's ``Integer.rotateLeft``. With the patch applied, it takes 3 seconds instead of 30 seconds. A 10x performance improvement.

I suspect that previously the JVM was not inlining ``rotateLeft`` for some reason, but with the patch applied, it does. With our scrypt parameters, ``Integers.rotateLeft`` is called 2^25 times, so it makes sense that this starts to add up.

For now we're keeping a copy of the ``SCrypt`` and ``Salsa20Engine`` classes in our codebase with this patch applied, but I figured I'd submit the change here as well to your thoughts on whether this is something that should be upstreamed.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/engines/Salsa20Engine.java
+++ b/core/src/main/java/org/bouncycastle/crypto/engines/Salsa20Engine.java
@@ -9,7 +9,6 @@ import org.bouncycastle.crypto.SkippingStreamCipher;
 import org.bouncycastle.crypto.constraints.DefaultServiceProperties;
 import org.bouncycastle.crypto.params.KeyParameter;
 import org.bouncycastle.crypto.params.ParametersWithIV;
-import org.bouncycastle.util.Integers;
 import org.bouncycastle.util.Pack;
 import org.bouncycastle.util.Strings;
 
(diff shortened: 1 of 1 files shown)


peterdettman:
Did you only see this improvement in a debug build? It appears from: https://github.com/beemdevelopment/Aegis/blob/5dfdbabf30be3ec87c9f72e2da51634c7c50388b/app/build.gradle#L85

that Aegis only uses proguard-android-optimize.txt in release builds.

If you can't also see the dramatic improvement in a release build, then perhaps the user's issue is yet something else.

alexbakker:
It's a bit of an odd situation. There are two places we've seen this:
- [This report](https://github.com/beemdevelopment/Aegis/issues/1024) of a user encountering this on a release build.
- Personally I can only reproduce this while running a debug build on an Android 13 device with a debugger attached. Without a debugger attached, the issue does not appear (at least for me), even on the debug build.

Anyway, looks like I may have jumped the gun a little bit here. Looking at the generated code in release builds, Proguard already removes the ``Integers.rotateLeft`` wrapper and replaces it with Java's ``Integer.rotateLeft``. I got confused because the flame graphs of the reporting user and mine looked so similar.

I'll close this for now. Apologies for the noise.

Issue #1297, opened 2022-12-06
Using custom DH groups


ottoka:
I recently noticed that bc-java can parse the _java.security_ property for custom DH groups (`jdk.tls.server.defaultDHEParameters`), but then does not use it in DHE-based TLS connections. The feature may be useful to anyone who needs to connect to legacy devices or in environments with prescribed DH parameters. I created a little patch for `ProvTlsServer` that fixes this, and will provide it as a PR (see #1298). Let me know what you think, and if you need me to change anything.

peterdettman:
Resolved by #1298 .

PR #1298, opened 2022-12-06
ProvTlsServer should use custom DH groups when configured

ottoka:
This adds support for using custom DH groups in TLS connections, see issue #1297 

1. It sorts the custom groups from `jdk.tls.server.defaultDHEParameters` property by bit length of P, to have an easier time selecting one later.
2. If such custom groups are configured, it picks the first one of appropriate bit length, honoring the `jdk.tls.ephemeralDHKeySize` property.

Regarding the associated TODOs I found in the code:
- TODO[jsse] Does this selection override the restriction from 'jdk.tls.ephemeralDHKeySize'?
  - It does not ovveride the property, but may include groups of smaller size, so the selection must honor the property.
- TODO[fips] Probably should be ignored in fips mode?
  - I cannot tell if this should be so or not, but if need you can easily add another condition and treat it as if the property were not set at all.


ottoka:
Fixed checkstyle issues in latest push.

PR #1310, opened 2022-12-28
missing type added to CertificateRequestMessageBuilder

kuehne-trustable-de:
Although the CertReqMsg class supports the AttributeTypeAndValue array, it is missing in the CertificateRequestMessageBuilder.
This PR adds the option to provide an 'AttributeTypeAndValue' array to the CertificateRequestMessageBuilder.


dghgit:
Thanks for the suggestion, merged with minor revision.

PR #1313, opened 2023-01-03
PoC for separate session keys for password based encryption

pschichtel:
While implementing my [pgcrypto-kt](https://github.com/pschichtel/pgcrypto-kt) library (a Kotlin reimplementation of the pgcrypto postgres extension for client side encryption) I faced 2 problems:

1. implementing the `sess-key` option was not directly¹ possible, since PBE encryption method does not trigger session info generation
2. implementing the `s2k-cipher-algo` option was not directly¹ possible, since the PBE encryption method will always use the cipher algo used for data.

¹: I eventually found dirty workarounds to trick BC into allowing it anyway, as posted on mailing list.

This PR extends the `PGPKeyEncryptionMethodGenerator` interface by the `wantsSessionInfo(): boolean` method, which is default-implemented to return true. The base class `PBEKeyEncryptionMethodGenerator` overrides that method and makes it configurable, while defaulting to false. Unless I'm missing something here, I'd argue this should be compatible even with user customizations of the method generators. This covers problem 1.

For problem 2 I added `getSessionInfoAlgo(): Integer` to `PGPKeyEncryptionMethodGenerator`, also default-implemented to return null. `PBEKeyEncryptionMethodGenerator` and `PublicKeyKeyEncryptionMethodGenerator` make it configurable. If the method returns a non-null value, it will be used to encrypt the session key. That covers problem 2. 

Method names are definitely open for discussion and I left out docs for the moment, until the actual code changes are accepted.


pschichtel:
@dghgit did you have a chance to look at this?

dghgit:
I've added a setForceSessionKey method for the forcing of the session key. That should show up here on github in a minute.

Now looking at the algorithm customisation

dghgit:
Okay, so for the last one I added setSessionKeyWrapperAlgorithm() to PBEKeyEncryptionMethodGenerator. Although I did end up with fairly minimal changes, doing this was cleanly was a bit more complicated than I thought it would be, I can see how you ended up where you did. The updated code for this should appear on github shortly.

End result:
- for 1: PGPEncryptedDataGenerator.setForceSessionKey(true)
- for 2: PBEKeyEncryptionMethodGenerator.setSessionKeyWrapperAlgorithm(PGPEncryptedData.AES_256)

If I've understood correctly this should both items in this PR. Let me know.

pschichtel:
@dghgit looks good to me, thanks for working on this, especially since it's a bit of a niche request! I can see why you went for more minimal changes.

PR #1325, opened 2023-01-30
#1324 pkix/bugfix/est npe on chunked transfer encoding

### LINKED ISSUE #1324: pkix/est: NullPointerException on HTTP/1.1 Transfer-Encoding chunked during ESTResponse processing
## Problem Statement

I am using EST simpleEnroll (bcpkix-jdk18on, 1.72) against an EST Server which answers the request using Transfer-Encoding Chunked. When the response is processed, the following exception occurs:

```
Exception in thread "main" org.bouncycastle.est.ESTException: Cannot invoke "java.lang.Long.longValue()" because "this.max" is null HTTP Status Code: 0
	at org.bouncycastle.est.ESTService.enroll(ESTService.java:322)
	at org.bouncycastle.est.ESTService.simpleEnroll(ESTService.java:352)
	at com.bbraun.cit.pki.est.Main.main(Main.java:59)
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because "this.max" is null
	at org.bouncycastle.est.CTEBase64InputStream.pullFromSrc(CTEBase64InputStream.java:43)
	at org.bouncycastle.est.CTEBase64InputStream.read(CTEBase64InputStream.java:104)
	at java.base/java.io.FilterInputStream.read(FilterInputStream.java:82)
	at org.bouncycastle.asn1.ASN1InputStream.readObject(ASN1InputStream.java:189)
	at org.bouncycastle.est.ESTService.handleEnrollResponse(ESTService.java:634)
	at org.bouncycastle.est.ESTService.enroll(ESTService.java:311)
	... 2 more
```

## Problem Analysis

The problematic code starts in [ESTResponse](https://github.com/bcgit/bc-java/blob/c8ff5b11a5a3cf643166f7f24d9f770e56041dc4/pkix/src/main/java/org/bouncycastle/est/ESTResponse.java):
https://github.com/bcgit/bc-java/blob/c8ff5b11a5a3cf643166f7f24d9f770e56041dc4/pkix/src/main/java/org/bouncycastle/est/ESTResponse.java#L165-L168

When we have a chunked transfer, `getContentLength()` returns `null` since no content length is known before the transfer is complete. So [CTEBase64InputStream's](https://github.com/bcgit/bc-java/blob/c8ff5b11a5a3cf643166f7f24d9f770e56041dc4/pkix/src/main/java/org/bouncycastle/est/CTEBase64InputStream.java) constructor receives `null` as its reading limit `max`. This leads to the fact that, `CTEBase64InputStream#pullFromSrc` produces the above mentioned NPE on the following line:

https://github.com/bcgit/bc-java/blob/c8ff5b11a5a3cf643166f7f24d9f770e56041dc4/pkix/src/main/java/org/bouncycastle/est/CTEBase64InputStream.java#L43-L46

## Solution

Passing a reading limit is not feasible when dealing with chunked transfers since the Content-Length header is not set. Besides chunked transfers, passing a reading limit is not neccessray at all since all inputStreams which are passed to the constructor return -1 when finished. All CTEBase64InputStream's methods respect this convention and consider the processing to be finished. Hence, removing this reading limit is fine.

I created a patch and tested it against real world servers. Chunked Transfers and non chunked transfers are working. 

I am happy to provide a PR.

adippel:
Fixes #1324

### DIFF
--- a/pkix/src/main/java/org/bouncycastle/est/CTEBase64InputStream.java
+++ b/pkix/src/main/java/org/bouncycastle/est/CTEBase64InputStream.java
@@ -14,13 +14,12 @@ class CTEBase64InputStream
     protected final byte[] rawBuf = new byte[1024];
     protected final byte[] data = new byte[768];
     protected final OutputStream dataOutputStream;
-    protected final Long max;
     protected int rp;
     protected int wp;
     protected boolean end;
     protected long read;
 
-    public CTEBase64InputStream(InputStream src, Long limit)
+    public CTEBase64InputStream(InputStream src)
     {
         this.src = src;
         this.dataOutputStream = new OutputStream()
@@ -31,7 +30,6 @@ class CTEBase64InputStream
                 data[wp++] = (byte)b;
             }
         };
-        this.max = limit;
     }
 
     // Pulls a line from the source, decodes it and returns the decoded length.
@@ -39,11 +37,6 @@ class CTEBase64InputStream
     protected int pullFromSrc()
         throws IOException
     {
-
-        if (this.read >= this.max)
-        {
-            return -1;
-        }
 
         int j = 0;
         int c = 0;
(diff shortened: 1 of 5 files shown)


dghgit:

PR #1345, opened 2023-03-05
OpenPGPv5 AEAD: Add missing out.close() call when closing AEADOutputS…

vanitasvitae:
…tream

I noticed that the AEADOutputStream classes do not close the underlying packet stream when getting closed. This results in truncated output packets.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcUtil.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcUtil.java
@@ -466,6 +466,7 @@ class BcUtil
             throws IOException
         {
             finish();
+            out.close();
         }
 
         private void writeBlock()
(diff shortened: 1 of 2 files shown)


dghgit:
I think this might cause issues - would a call to flush() fix the problem you are seeing?

vanitasvitae:
Calling flush is not sufficient in my experience.

See the test I added (please do not merge the test since it is adapted from #1346).
If you revert https://github.com/bcgit/bc-java/pull/1345/commits/a56b42220bb98b503bd069a99cde52bd52cb4f36 you will see that the test starts to fail.

vanitasvitae:
Is my test case perhaps misusing the API?

dghgit:
No, you're actually correct - the encryptor classes are only used with ClosableBCPGInputStream objects - in this case the close method is actually used to finish the stream in the layer above but it doesn't propagate the close further meaning further PGP packets can be written to the stream as required. Being a while since I looked at this, a lot of code has flooded under the bridge since then. Merged with minor changes. Should appear on github shortly. Thanks!

PR #1362, opened 2023-03-26
typo fixed in returned string by getAlgorithmName() in XoodyakDigest

jonelo:
in XoodyakDigest.java, method getAlgorithmName(), the algorithm name should be "Xoodyak Hash", not "Xoodak Hash".


dghgit:

Issue #1363, opened 2023-03-26
ArrayIndexOutOfBoundsException at System.arraycopy in Blake2spDigest update method for message length 1007 bytes


jonelo:
When hashing a message length of 1007 bytes with Blake2spDigest I got the following exception:
```
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 495 out of bounds for byte[256]
```
I figured out that in that case the arraycopy call failed at line 102 at
```
            System.arraycopy(message, offset, buffer, left, len);
```
because of the values offset=512, buffer.length=256, left=0, and len=495

jonelo:
a possible fix would be to increase the buffer size to 512 bytes as in commit https://github.com/jonelo/bc-java/commit/cdaf22c87886aeac38b8afb6410ebefb3cd6443d

dghgit:
Fixed in 1.73

PR #1364, opened 2023-03-26
Increase buffer size in Blake2spDigest in order to avoid an ArrayIndexOutOfBoundsException

jonelo:
possible fix for issue #1363


dghgit:
Thanks for the patch! Merged with minor revision.

Issue #1379, opened 2023-04-20
PGPSignature verification is not thread-safe


vanitasvitae:
A user recently noticed, that some PGP operations (decryption of a message in this case) would sometimes fail randomly.
It turns out, their code was multi-threaded, and PGPainless would prior to message decryption check binding signatures on the encryption subkey.

This caused the following error in BCs PGPSignature class:
```
java.lang.ArrayIndexOutOfBoundsException: length=8; index=8
at org.bouncycastle.crypto.digests.LongDigest.update(Unknown Source:8)
at org.bouncycastle.crypto.digests.LongDigest.update(Unknown Source:8)
at org.bouncycastle.openpgp.operator.bc.BcImplProvider$EdDsaSigner.update(Unknown Source:2)
at org.bouncycastle.openpgp.operator.bc.SignerOutputStream.write(Unknown Source:2)
at org.bouncycastle.openpgp.PGPSignature.blockUpdate(Unknown Source:2)
at org.bouncycastle.openpgp.PGPSignature.update(Unknown Source:16)
at org.bouncycastle.openpgp.PGPSignature.update(Unknown Source:2)
at org.bouncycastle.openpgp.PGPSignature.updateWithPublicKey(Unknown Source:21)
at org.bouncycastle.openpgp.PGPSignature.verifyCertification(Unknown Source:4)
```

After some digging, we figured out, that this is probably caused by multiple threads working on the same PGPPublicKeyRing instance (and therefore on the same PGPSignature instance, hence the same SignerOutputStream using the same EdDsaSigner, using the same LongDigest).

Upon further inspection, it turns out that PGPSignature is not designed to be used in a multi-threaded environment. The SignerOutputStream is a member of the class, so parallel verification would probably write garbage to that stream, resulting in verification fails.

Would you consider a patch that refactors the PGPSignature class in such a way that the stateful verification code is moved into a separate class, even though such a patch would break the API backwards compatibility?

I could imagine something along these lines:
```java
PGPSignature signature = ...;
SignatureVerifier verifier = signature.init(verifierBuilderProvider, pubKey); // stateful stuff is moved here
boolean valid = verifier.verifyCertification(masterKey, pubKey);
```

I think this would allow multiple threads to work on the same PGPSignature instance without interfering (since each thread would have its own copy of the SignatureVerifier class).

Now that I think about it, if instead of refactoring the `init` method, we add a separate method (`betterInit`?), we can even change the behavior without breaking backwards compat.

What do you think?

[...]

PR #1383, opened 2023-04-25
Separate PGPSignature and its verification state

vanitasvitae:
As discussed in #1379, attempting to do multi-threaded operations on keys might result in race conditions during parallel signature verification of the same signature.

This PR separates out the state of a signature verification and allows to acquire per-thread verification objects for each signature.

This PR is fully backwards compatible.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
@@ -52,9 +52,8 @@ public class PGPSignature
     private SignaturePacket sigPck;
     private int signatureType;
     private TrustPacket trustPck;
-    private PGPContentVerifier verifier;
-    private byte lastb;
-    private OutputStream sigOut;
+
+    private Verification verification;
 
     private static SignaturePacket cast(Packet packet)
         throws IOException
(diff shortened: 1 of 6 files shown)


dghgit:
So yes, signatures and ciphers for that matter are inherently stateful and not thread safe.

I'm sympathetic, but I'll have to think about this one, the verification class isn't really thread safe either, it's more that a different object is getting created rather than another PGPSignature - perhaps a better approach would be to use a ThreadLocal internally or accept the fact that an operation like this is inherently stateful. Thinking...

dghgit:
Okay, I've had a bit of a sleep, folded a few cranes, pottered around, and stared out the window a bit. I don't think this is the right approach, none of these things in Java are thread safe and it'll really break the API as well, without really fixing the problem.

The problem, as outlined by the use case, is kind of interesting though, as in for what's described, it should be possible to do this, and it would be generally useful. I'd like to suggest we still add a verifier class, but as a new API class in it's own right, but make it properly thread safe and suitable for small messages and certifications, more a straight "here's the message, here's the sig - yes or no", so no update methods, no room for confusion or misuse, no risk of internal state. What do you think?

We verify things a lot more times than we sign them, and it does seem obvious that in the interests of "liveliness" there would be a class that would make doing this in a thread safe way possible.

vanitasvitae:
> The problem, as outlined by the use case, is kind of interesting though, as in for what's described, it should be possible to do this, and it would be generally useful. I'd like to suggest we still add a verifier class, but as a new API class in it's own right, but make it properly thread safe and suitable for small messages and certifications, more a straight "here's the message, here's the sig - yes or no", so no update methods, no room for confusion or misuse, no risk of internal state. What do you think?

So the new API would be intended primarily for signatures over small pieces of data, e.g. binding signatures on keys, but not message-signatures, right?

dghgit:
Yep, that's what I'm thinking.

vanitasvitae:
In https://github.com/bcgit/bc-java/pull/1383/commits/f426a2be8e7976efb9d5108979180461f652acfc I prototyped an (untested!) API for signature verification.

Does this go into the rough direction you have in mind?
Example:

```java
PGPPublicKey primaryKey = ...;
PGPSignature userIdSig = ...;

PGPContentVerifierBuilderProvider verifierBuilderProvider = new BcPGPContentVerifierBuilderProvider();
PGPSignatureVerifierBuilder verifierBuilder = new PGPSignatureVerifierBuilder(userIdSig, primaryKey, verifierBuilderProvider);

PGPSignatureVerifier verifier = verifierBuilder.userIdSignature("Alice <alice@example.org>", primaryKey);
assert(verifier.verify());
```

vanitasvitae:
In PGPainless, I have a rather advanced set of validity checkers for signatures in place, which check additional conditions for signature validity, like the existence of a creation time subpacket, critical unknown subpackets/notations and so on (some [test results](https://tests.sequoia-pgp.org/#Detached_signature_with_Subpackets)).
PGPainless also not only verifies whether a signature is correct, but also whether the certificate chain from signature over signing key to primary key is valid (no expired/revoked keys etc.).

I wrote some blog posts about signature verification with OpenPGP, which illustrate the challenge:
* https://blog.jabberhead.tk/2021/04/03/why-signature-verification-in-openpgp-is-hard/
* https://blog.jabberhead.tk/2022/09/14/using-pushdown-automata-to-verify-packet-sequences/
* https://blog.jabberhead.tk/2022/10/26/implementing-packet-sequence-validation-using-pushdown-automata/

Would you, for a new signature verification API, be interested in me attempting to upstream some or all these checks?

dghgit:
I've had a go. I left a PGPSignatureVerifier in as it's useful to have class carrying most of the computation load when you're trying to distribute things. The PGPSignatureVerifierBuilder just creates verifying objects. I've tried to minimize duplication.

The kind of use I'm imagining is being able to create a set of signatures with their corresponding verifiers, but then only invoking the verifiers on the signatures you care about. 

Anyway, it should be on github shortly, let me know what you think.

vanitasvitae:
Your changes look good!

dghgit:
Great. One last one, in case I haven't replied on this elsewhere, yes would be interested adding anything which would improve peoples ability to validate signatures more effectively.

hermanwongkm:
Is this issue fixed? Was it merged? I seem to be getting the same error with the latest version of pgppainless.

PR #1384, opened 2023-04-26
Add support for wildcard keyIDs when generating OpenPGP messages

vanitasvitae:
This PR allows the user to generate encrypted messages, where the recipient key-id of a public-key-encrypted session key packet is replaced by a wildcard (0).

This can simply be accomplished like so:
```java
        ...
        PublicKeyKeyEncryptionMethodGenerator pkeMethodGen = new BcPublicKeyKeyEncryptionMethodGenerator(
                recipientPubKey);
        pkeMethodGen.setUseWildcardKeyID(true);
        encDataGen.addMethod(pkeMethodGen);
        ...
```

Since this PR merely adds a new method, it is fully backwards compatible.


vanitasvitae:
Motivation for this patch is that sending encrypted messages with BCC'd recipients would leak who the recipients are via recipient key-IDs in PKESK packets.

dghgit:
Thanks for the patch. Now merged, will appear in 174b01.

Issue #1390, opened 2023-05-08
JcaJce: Cannot find any provider supporting AES/OCB/NoPadding or AES/EAX/NoPadding


vanitasvitae:
Hey!
I'm currently testing SEIPDv2 support with Bouncycastles beta release 174b01.

I noticed, that decryption of AEAD encrypted messages using EAX and OCB fails when using the JcaJce API.

```
org.bouncycastle.openpgp.PGPException: cannot create cipher: Cannot find any provider supporting AES/EAX/NoPadding
	at org.bouncycastle.openpgp.operator.jcajce.OperatorHelper.createCipher(Unknown Source)
	at org.bouncycastle.openpgp.operator.jcajce.JceAEADUtil.createAEADCipher(Unknown Source)
	at org.bouncycastle.openpgp.operator.jcajce.JcePBEDataDecryptorFactoryBuilder$1.recoverAEADEncryptedSessionData(Unknown Source)
	at org.bouncycastle.openpgp.PGPPBEEncryptedData.getSessionKey(Unknown Source)
	at org.bouncycastle.openpgp.PGPPBEEncryptedData.getDataStream(Unknown Source)
```

The stacktrace for OCB looks similar, just the algorithm is swapped out.

The exceptions do not happen if I use a `BcPBEDataDecryptorFactory` instead.

Is this to be expected?

vanitasvitae:
Nevermind, this issue appears to be my faulty setup.

vanitasvitae:
Reopening, since this is actually a bug.

dghgit:
I assume #1391 is the fix for this one?

vanitasvitae:
Correct :)

PR #1391, opened 2023-05-09
JcePBEDataDecryptorFactoryBuilder.setProvider(): Inject provider into…

vanitasvitae:
… JceAEADUtil instance

Consider a setup, where the systems SecurityProviders are not changed (i.e. there is no `Security.addProvider(new BouncyCastleProvider())` call).

It is still possible to use different providers in the Jce classes by calling `setProvider()`.
This method injects the security provider instance, so that it gets used to instantiate Ciphers etc.

In case of the `JcePBEDataDecryptorFactoryBuilder, there are two places where the provider needs to be changed when `setProvider()` is called, as code depends on it: 
* `helper` is an `OperatorHelper` that gets initialized with a `ProviderJcaJceHelper` with the given provider
* `aeadHelper` uses `helper` to provide AEAD-related functionality.

Now, when `setProvider()` is called, the `helper` is replaced with a fresh `ProviderJcaJceHelper` instance, but the `aeadHelper` still references the old `OperatorHelper` instance, which by default is a `DefaultJcaJceHelper` object.

As a result, while the user has called `setProvider()`, some OpenPGP operations using AEAD ciphers (e.g. AES/OCB/NoPadding) will fail (see https://github.com/bcgit/bc-java/issues/1390), since the JceAEADUtil `aeadHelper` uses the `DefaultJcaJceHelper` which cannot utilize the set SecurityProvider.

The patch fixes this by replacing the `JceAEADUtil` instance too when calling `setProvider()`.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcePBEDataDecryptorFactoryBuilder.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcePBEDataDecryptorFactoryBuilder.java
@@ -64,6 +64,7 @@ public class JcePBEDataDecryptorFactoryBuilder
     public JcePBEDataDecryptorFactoryBuilder setProvider(Provider provider)
     {
         this.helper = new OperatorHelper(new ProviderJcaJceHelper(provider));
+        this.aeadHelper = new JceAEADUtil(helper);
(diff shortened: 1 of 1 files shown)


dghgit:
Thanks! Merged.

PR #1394, opened 2023-05-14
Update examples

vanitasvitae:
I noticed that the example classes in the OpenPGP module are quite out of date and do not meet modern security standards anymore (e.g. 1024 bit RSA keys etc.).

So this PR not only modernizes some example classes (3072 bit RSA keys, 3072 bit ElGamal keys), but also adds a new EllipticCurveKeyPairGenerator example that generates modern OpenPGP v4 keys based on Curve25519.

Further, it properly initializes OpenPGP keys with signature subpackets containing proper key flags and modern algorithm preferences.


dghgit:
Thanks for the patch! Merged.

PR #1406, opened 2023-05-26
OpenPGP: Switch on integrity protection by default

vanitasvitae:
Greetings from the 7th. OpenPGP Summit in Geneva!

This PR enables integrity protection for OpenPGP messages by default. It can of course by disabled by calling `dataEncryptorBuilder.setWithIntegrityProtection(false)`.

Generating messages without integrity protection has been discouraged for a while now, so I thought it is about time.


dghgit:
Agreed, merged.

PR #1408, opened 2023-05-31
Specify UTF-8 encoding in bc+-build.xml

scottarc:
If the default is set to ASCII, this fails to build.


dghgit:

PR #1411, opened 2023-06-02
OpenPGP v6 elliptic curve key formats

vanitasvitae:
This PR adds support for parsing new OpenPGP v6 key algorithms X25519, X448, Ed25519, Ed448.

Note, that while some of these algorithms were already sort of supported in OpenPGP v4 (`EDDSA_LEGACY`), v6 properly introduces them with a changed encoding wire format: https://openpgp-wg.gitlab.io/rfc4880bis/#name-public-key-algorithms

To prevent duplicated code, I also introduced `OctetArrayPublicBCPGKey` and `OctetArraySecretBCPGKey` classes, which handle the byte array length enforcing etc.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/Ed25519PublicBCPGKey.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/Ed25519PublicBCPGKey.java
@@ -1,0 +1,20 @@
+package org.bouncycastle.bcpg;
+
+import java.io.IOException;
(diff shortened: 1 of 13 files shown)


vanitasvitae:
I split this PR out of https://github.com/bcgit/bc-java/pull/1353 for better readability.

dghgit:
Thanks for the patch! Now merged.

PR #1412, opened 2023-06-02
Backport latest changes to the PKESK v6 packet

vanitasvitae:
See https://gitlab.com/openpgp-wg/rfc4880bis/-/merge_requests/304
The change to the format allows to parse arbitrary PKESKv6 packets, even if the key version, and hence the fingerprint length is unknown.


dghgit:
Thanks for the patch. These changes have also been incorporated into 174b05 (should appear on github shortly as well).

PR #1413, opened 2023-06-02
V6 fingerprint calculation

vanitasvitae:
This PR adds support for calculating fingerprints of v6 keys.

For the test to run successfully, merging #1411 is required.


dghgit:
Thanks for the patch, now merged as well. One extra note this is also up in 174b05 on https://www.bouncycastle.org/betas - we're releasing 1.74 early so 174b05 is really the release candidate, if you could try it out and let me know if there are any issues. Thanks.

vanitasvitae:
I just tested the beta and found one issue that I will shortly provide another patch for (unrelated to this issue).

vanitasvitae:
Nevermind, it was just a problem on my end. Looks all fine now :)

PR #1422, opened 2023-06-13
Implement OpenPGP Packet Criticality

vanitasvitae:
The upcoming OpenPGP v6 specification introduces the concept of [Packet Criticality](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-09.html#packet-criticality) based on packet tags.

This PR adds support for packet criticality by adding `getPacketTag()` to the `Packet` class.
Further, it introduces a dedicated `UnknownPacket` class to improve [forwards compatibility](https://mailarchive.ietf.org/arch/msg/openpgp/QUiEKx3PQeJOXnkcvvnuHpv739M/) with future revisions of the standard, as well as a dedicated `ReservedPacket` class.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/AEADEncDataPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/AEADEncDataPacket.java
@@ -118,4 +118,10 @@ public class AEADEncDataPacket
     {
         return AEADUtils.getIVLength(aeadAlgorithm);
     }
+
+    @Override
+    public int getPacketTag()
+    {
+        return getType();
+    }
 }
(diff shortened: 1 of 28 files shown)


vanitasvitae:
OpenPGP v6 implementations would set `pgpObjectFactory.setThrowForUnknownCriticalPackets(true);`, such that unknown critical packets automatically cause an `IOException` to be thrown.
If this flag is not set, the object factory will instead return an `UnknownPacket` object which the user can further consume.

dghgit:
Thanks for the patch. Merged with some changes (trying to preserve API compatibility). This is also now published on https://www.bouncycastle.org/betas 176b05 or later.

PR #1423, opened 2023-06-13
ArmoredInputStream: Make CRC calculation optional

vanitasvitae:
The [crypto-refresh-09 document states](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-09.html#name-optional-checksum) that the CRC sum of ASCII armor is considered optional.

Really, if some new packet formats are used (SEIPDv2), the ASCII armored output MUST NOT even include a CRC sum:

> The CRC24 footer MUST NOT be generated if it can be determined by context or by the OpenPGP object being encoded that the consuming implementation accepts base64 encoded blocks without CRC24 footer. Notably:
> * An ASCII-armored Encrypted Message packet sequence that ends in an v2 SEIPD packet MUST NOT contain a CRC24 footer.
> * An ASCII-armored sequence of Signature packets that only includes v6 Signature packets MUST NOT contain a CRC24 footer.
> * An ASCII-armored Transferable Public Key packet sequence of a v6 key MUST NOT contain a CRC24 footer.
> * An ASCII-armored keyring consisting of only v6 keys MUST NOT contain a CRC24 footer.

This PR allows the user to configure `ArmoredOutputStream` objects to not calculate and include a CRC sum using `noCRCSum()`.
It also allows the user to disable CRC checksum calculation altogether when consuming armored data using and `ArmoredInputStream` by calling `ignoreCRCSum()`.

Some tests are included as well.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
@@ -179,6 +179,18 @@ public class ArmoredInputStream
 
         start = false;
     }
+
+    /**
+     * Disable calculation of the CRC sum.
+     * If the CRC check is disabled, missing or malformed CRC sums are ignored.
+     *
+     * @return armored input stream
+     */
+    public ArmoredInputStream ignoreCRCSum()
+    {
+        this.crc = null;
+        return this;
+    }
     
     public int available()
         throws IOException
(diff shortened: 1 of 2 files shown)


vanitasvitae:
I reverted the changes to the `ArmoredOutputStream` in favor of #1425.

dghgit:
Thanks! Merged with revision (added Builder).

PR #1424, opened 2023-06-13
ArmoredOutputStream: Do not include a Version: header by default

vanitasvitae:
The [crypto-refresh-09 document states](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-09.html#section-6.2.2.1) that in order to minimize metadata, implementations SHOULD NOT include a `Version:` header when forming ASCII armor.
This PR changes BCs default behavior to not include a version header by default.


dghgit:
Given the other PR, I think we need to add a Builder to the ArmoredOutputStream which allows these things to be configured, not sure what to do about input stream yet. Suggest dropping OutputStream commits from #1423 and adding the builder to this one that covers making the two items optional.

vanitasvitae:
Good point! I will rework the PR :)

vanitasvitae:
Closed in favour of #1425

PR #1425, opened 2023-06-14
Rework the ArmoredOutputStream

vanitasvitae:
This PR introduces a builder for the ArmoredOutputStream, allowing configuration of armor headers and whether to include a CRC checksum into the armored output.

This PR invalidates #1424 and parts of #1423


dghgit:
Thanks! Merged with some revisions, I dropped HashAlgorithmUtils as the names are really tied to the ArmoredOutputStream - as soon as we publish HashAlgorithmUtils people will start using it for other things and then get upset if we change it. I've also changed it stay as it was if the old way of building the AmoredOutputStream is used, I think this is probably better as it means people relying on the presentation for tests and the like won't find things have changed unless they make the change themselves.

In other news, master on bc-java is now called "main" - see discussion for details.

PR #1426, opened 2023-06-14
Document S2K types and usages and mark deprecated values accordingly

vanitasvitae:
This PR adds documentation about weaknesses of different S2K types and usages and marks insecure options as deprecated.


dghgit:
Thanks. Merged!

PR #1427, opened 2023-06-14
Properly parse v6 secret keys and implement unlocking AEAD protected keys

vanitasvitae:


vanitasvitae:
I'm still working on AEAD encryption for secret keys, which probably makes sense to be part of this PR as well.

PR #1463, opened 2023-08-01
PKCS1Encoding should use short-circuit AND operator

johnshajiang:
It may be better to use the logical operator `&&` rather than `&` in `org.bouncycastle.crypto.encodings.PKCS1Encoding`.


dghgit:
It sort of depends on the definition of better. In this case the slow "and" is used as it improves the chance of the process being constant time -  important in this case.

johnshajiang:
Get it, thanks!

PR #1410, opened 2023-06-02
Add support for crypto-refresh style PreferredAEADCiphersuites signat…

vanitasvitae:
…ure subpacket

This PR adds support for the Preferred AEAD Ciphersuites subpacket [as defined](https://openpgp-wg.gitlab.io/rfc4880bis/#name-preferred-aead-ciphersuites) by the crypto-refresh document.

This packet lists supported combinations of symmetric algorithms and AEAD algorithms.

PR #1464, opened 2023-08-01
Properly parse and set PreferredAEADCipherSuites subpacket

vanitasvitae:
In https://github.com/bcgit/bc-java/pull/1410 I apparently forgot to change some places away from PreferredAlgorithms to PreferredAEADCiphersuites.

This PR fixes these cases.


dghgit:
Would you deprecate the current methods and add setPreferredAEADCiphersuites and getPreferredAEADCiphersuites instead? Thanks to the latest GPG role out a lot of people seem to be using AEAD now - I think it's too late to break the API on this one, there's going to be a lot of third party code depending on the current methods.

vanitasvitae:
Sure, I can change that :)

vanitasvitae:
Let me add some more tests for this, so that we can be sure that both methods are not interfering with another.

vanitasvitae:
Closed in favor of #1743

PR #1471, opened 2023-08-15
Extend the HPKE API

thomdev:
Extend the HPKE API to:
 - Support encryption/decryption from byte ranges, in order to avoid copies.
 - Allow the sender to specify their ephemeral key pair, which is necessary to validate OHTTP test vectors.
 - Expose non-labeled extract and expand operations on the HKDF.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/hpke/AEAD.java
+++ b/core/src/main/java/org/bouncycastle/crypto/hpke/AEAD.java
@@ -51,7 +51,7 @@ public class AEAD
             throw new IndexOutOfBoundsException("Invalid offset");
         }
         if (ptOffset + ptLength > pt.length) {
-            throw new IllegalStateException("Invalid length");
+            throw new IndexOutOfBoundsException("Invalid length");
         }
 
         CipherParameters params;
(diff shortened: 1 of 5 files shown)


normanmaurer:
Thanks a lot @thomdev for the PR... We would love to have this change so we could eventually create an [OHTTP](https://www.ietf.org/archive/id/draft-ietf-ohai-ohttp-09.html) codec for [netty](https://netty.io).

core/src/main/java/org/bouncycastle/crypto/hpke/AEAD.java
```diff
@@ -61,19 +68,32 @@ public byte[] seal(byte[] aad, byte[] pt)
...
+        return this.seal(aad, pt, 0, pt.length);
+    }
 
     // used by Receiver
-    public byte[] open(byte[] aad, byte[] ct)
+    public byte[] open(byte[] aad, byte[] ct, int ctOffset, int ctLength)
         throws InvalidCipherTextException
     {
+        if (ctOffset < 0 || ctOffset > ct.length) {
+            throw new IllegalStateException("Invalid offset");
```
normanmaurer, line 91:
```suggestion
            throw new IndexOutOfBoundsException("Invalid offset");
```

normanmaurer (changes requested):

dghgit:

PR #1480, opened 2023-08-24
IETFUtils.rDNsFromString: throw exception if unescaped equal sign is used in value

ff-wl:
If a user calls IETFUtils.rDNsFromString and forgets to escape an equal sign in a value, the remaining part is omitted silently.
e.g: 
'CN=foo=bar' results in a X500Name with CN = 'foo'

With this change an IllegalArgumentException is thrown if the equal sign is not escaped:

'CN=foo=bar' -> throw IllegalArgumentException
'CN=foo\\=bar' -> results in CN = 'foo=bar'


dghgit:

Issue #1489, opened 2023-09-14
Bouncycastle does not retransmit the last server handshake flight on receipt of the retransmitted last client flight


JonathanLennox:
When performing a DTLS handshake as the server, BouncyCastle does not respond to a retransmission of the client's second DTLS flight with a retransmission of the server flight.

See the attached PCAP (gzip'd because GitHub doesn't allow .pcap files).
[Stuck-DTLS-handshake.pcap.gz](https://github.com/bcgit/bc-java/files/12613385/Stuck-DTLS-handshake.pcap.gz)

This is a handshake between Chrome 117.0.5938.62 and BouncyCastle 1.75, with artificial packet reordering introduced with the Linux tc tool (`netem delay 100.0ms 75.0ms`).  A bug in Chrome's stack causes it not to handle the case where the packet containing the server's Encrypted "Finish" message precedes the packet containing the server's ChangeCipherSpec message, so Chrome retransmits its "Certificate, Client Key Exchange, Certificate Verify, Change Cipher Spec, Encrypted Handshake Message" flight (which happens to be aggregated into a single packet).  

This problem appears to have been introduced in BouncyCastle 1.74; BouncyCastle 1.73 properly retransmits the server flight.  1.76 shows the same behavior as 1.75.

Tested with Chrome (as mentioned) using WebRTC, talking to the Jitsi Videobridge which is using (the various versions of ) BouncyCastle-Java.

JonathanLennox:
It looks like the problem was introduced in commit 230783692cf5112ab04707241677fa2628accca3 by @peterdettman , and the issue is that `DTLSRecordLayer.receivePendingRecord` doesn't recognize the client's `change_cipher_spec` from the previous epoch as being part of the retransmitted handshake.

JonathanLennox:
More specifically, when `receivePendingRecord` fails to parse the `change_cipher_spec`, it clears out the entirety of the rest of `recordQueue`, which includes the encrypted `Finish" message. Thus the "Finish" message is never processed, and so the server flight is not retransmitted.

JonathanLennox:
A fix is now in https://github.com/bcgit/bc-java/pull/1491.

JonathanLennox:
[...]

PR #1491, opened 2023-09-18
Fix retransmission of DTLS server flight after a retransmitted aggregated packet

JonathanLennox:
This fixes the issue described in https://github.com/bcgit/bc-java/issues/1489, and adds unit tests.

The records parsed here, and handed to record processing, are still a subset of the records that were parsed prior to commit 2307836, so this should be safe.


JonathanLennox:
Note: I updated this PR to remove some files that got checked in accidentally (gradlew etc.)

peterdettman:
This PR has now been merged and will be included in the 1.77 release that we are currently preparing. We especially appreciate the extra test cases.

PR #1497, opened 2023-09-29
Turn on header parsing by default in ArmoredInputStream.Builder

vanitasvitae:
I noticed, that when constructing an `ArmoredInputStream` via the old constructor `ArmoredInputStream(inputStream)`, header parsing is enabled by default, while `ArmoredInputStream.builder().build(inputStream)` disables header parsing.

This PR brings the builder in line with the old constructor by enabling header parsing by default.


dghgit:
Yes, good point. Merged!

PR #1500, opened 2023-10-02
CMS: Add support for sha-1WithRSAEncryption *digest* algorithm + aliases.

kevinhwang:
This adds support to the CMS libraries for the `sha-1WithRSAEncryption` *digest* algorithm (OID `1.3.14.3.2.29`) and its aliases. This is not to be confused with the `SHA1withRSA` *signature* algorithm which is currently already supported.

Per [RFC 3279 2.2.1](https://datatracker.ietf.org/doc/html/rfc3279#section-2.2.1) and [RFC 8017 A.2.4](https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.4), this digest algorithm computes the SHA-1 digest on the encoded `DigestInfo` whose `digestAlgorithm` field is set to the `id-sha1` OID.

Bouncy Castle doesn't currently recognize this digest algorithm (either for building `CMSSignedData` or verifying) and fails with

```
org.bouncycastle.cms.CMSException: can't create digest calculator: exception on setup: java.security.NoSuchAlgorithmException: no such algorithm: 1.3.14.3.2.29 for provider BC
```

In versions 1.69 and earlier, recognition for OID `1.3.14.3.2.29` in `OperatorHelper.java` was present, but the data encoding was not performed correctly (setting OID `1.3.14.3.2.29` during the data encoding step, rather than OID `1.3.14.3.2.26` as RFC 3279 requires), so the computed digest would be wrong and not verify.

### DIFF
--- a/pkix/src/test/java/org/bouncycastle/cms/test/NewSignedDataTest.java
+++ b/pkix/src/test/java/org/bouncycastle/cms/test/NewSignedDataTest.java
@@ -18,6 +18,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import junit.framework.Assert;
@@ -1623,70 +1624,86 @@ public class NewSignedDataTest
         rsaPSSTest("SHA3-384withRSAandMGF1");
     }
 
-    public void testSHA1WithRSADigest()
+    public void testSHA1WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA1withRSA");
+        rsaSignatureTest("SHA1withRSA");
     }
 
-    public void testSHA224WithRSADigest()
-        throws Exception
+    public void testSHA256WithRSASignatureAndSHA1WithRSAEncryptionDigest()
+            throws Exception
     {
-        rsaDigestTest("SHA224withRSA");
+        rsaSignatureAndDigestTest("SHA256withRSA",
+            // sha-1WithRSAEncryption (RFC 3279 2.2.1)
+            Optional.of(new AlgorithmIdentifier(new ASN1ObjectIdentifier("1.3.14.3.2.29"))));
     }
 
-    public void testSHA256WithRSADigest()
-        throws Exception
+    public void testSHA256WithRSASignatureAndSHA1WithRSASignatureDigest()
+            throws Exception
     {
-        rsaDigestTest("SHA256withRSA");
+        rsaSignatureAndDigestTest("SHA256withRSA",
+                // sha-1WithRSASignature (RFC 8017 A.2.4)
+                Optional.of(new AlgorithmIdentifier(new ASN1ObjectIdentifier("1.2.840.113549.1.1.5"))));
     }
 
-    public void testSHA384WithRSADigest()
+    public void testSHA224WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA384withRSA");
+        rsaSignatureTest("SHA224withRSA");
     }
 
-    public void testSHA512WithRSADigest()
+    public void testSHA256WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA512withRSA");
+        rsaSignatureTest("SHA256withRSA");
     }
 
-    public void testSHA3_224WithRSADigest()
+    public void testSHA384WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA3-224withRSA");
+        rsaSignatureTest("SHA384withRSA");
     }
 
-    public void testSHA3_256WithRSADigest()
+    public void testSHA512WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA3-256withRSA");
+        rsaSignatureTest("SHA512withRSA");
     }
 
-    public void testSHA3_384WithRSADigest()
+    public void testSHA3_224WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA3-384withRSA");
+        rsaSignatureTest("SHA3-224withRSA");
     }
 
-    public void testSHA3_512WithRSADigest()
+    public void testSHA3_256WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA3-512withRSA");
+        rsaSignatureTest("SHA3-256withRSA");
     }
 
-    public void testSHA512_224ithRSADigest()
+    public void testSHA3_384WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA512(224)withRSA");
+        rsaSignatureTest("SHA3-384withRSA");
     }
 
-    public void testSHA512_256ithRSADigest()
+    public void testSHA3_512WithRSASignature()
         throws Exception
     {
-        rsaDigestTest("SHA512(256)withRSA");
+        rsaSignatureTest("SHA3-512withRSA");
+    }
+
+    public void testSHA512_224ithRSASignature()
+        throws Exception
+    {
+        rsaSignatureTest("SHA512(224)withRSA");
+    }
+
+    public void testSHA512_256ithRSASignature()
+        throws Exception
+    {
+        rsaSignatureTest("SHA512(256)withRSA");
     }
 
     public void testEd25519()
@@ -2207,8 +2224,14 @@ public class NewSignedDataTest
         verifySignatures(s, md.digest("Hello world!".getBytes()));
     }
 
-    private void rsaDigestTest(String signatureAlgorithmName)
+    private void rsaSignatureTest(String signatureAlgorithmName)
         throws Exception
+    {
+        rsaSignatureAndDigestTest(signatureAlgorithmName, Optional.empty());
+    }
+
+    private void rsaSignatureAndDigestTest(String signatureAlgorithmName, Optional<AlgorithmIdentifier> digestAlgorithm)
+            throws Exception
     {
         List certList = new ArrayList();
         CMSTypedData msg = new CMSProcessableByteArray("Hello world!".getBytes());
@@ -2225,6 +2248,8 @@ public class NewSignedDataTest
         JcaSignerInfoGeneratorBuilder siBuilder = new JcaSignerInfoGeneratorBuilder(new JcaDigestCalculatorProviderBuilder().setProvider(BC).build());
 
         gen.addSignerInfoGenerator(siBuilder.build(contentSigner, _origCert));
+
+        digestAlgorithm.ifPresent(siBuilder::setContentDigest);
 
         gen.addCertificates(certs);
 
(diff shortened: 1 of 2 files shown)


dghgit:
I've had a chance to have a quick look at this, essentially we'd be adding support for a specific kind of broken CMS signed data message. 

sha-1WithRSAEncryption is a signature algorithm, the reason for the nomenclature lies in the original drafts of PKCS#7 where the algorithm used for processing the hash was referred to an an "encryption algorithm" since the process of signing with RSA is the same as "encrypting with the private key". Neither of the OIDs being added to map to SHA-1 are, or ever have been, considered to be identifying the SHA-1 hash algorithm, both RFCs are explicit about requiring 1.3.14.3.2.26 as being the OID used to represent SHA-1. The move to calling it a signatureAlgorithm was a result of other signing algorithms being required which made it obvious that the RSA centric view of the universe was not always correct.

It doesn't look like this would present a security issue, but I would prefer to keep the number of non-compliant OIDs to a minimum. Are you saying that you have message where both the OIW and PKCS definitions for SHA-1withRSA are used to represent the SHA-1 OID? Or is it just one variant that is present (I'd guess the OIW arc) and the other is given "for completeness".

kevinhwang:
> Are you saying that you have message where both the OIW and PKCS definitions for SHA-1withRSA are used to represent the SHA-1 OID

Yup, unfortunately we have clients sending CMS messages where the the digest algorithm OID (what BouncyCastle's CMS library calls the "ContentDigestAlgorithm") is sha-1WithRSAEncryption, even though they really should be setting it as SHA1, and we're trying to support processing and verifying these messages.

dghgit:
We'd be better off if people just did it correctly. In some situations it is actually possible to weaken the security of the message when issues like this have to be dealt with. One last question, do they use sha-1WithRSAEncryption as the encryption algorithm in the signed data message or is it just rsaEncryption?

kevinhwang:
> One last question, do they use sha-1WithRSAEncryption as the encryption algorithm in the signed data message or is it just rsaEncryption?

Do you mean the signature algorithm? They appear to use rsaEncryption as the identifier. Here's an example of a partial ASN.1 dump:

image

dghgit:
So I've pushed an update to SignerInformation.java which should deal with this - in a limited fashion. Is it possible to provide a sample of one of these messages?

kevinhwang:
> So I've pushed an update to SignerInformation.java which should deal with this - in a limited fashion.

Thanks! It does seem like it allows decoding and verification if CMS messages using the weird digest algo name.

Unfortunately, it still doesn't allow you to sign messages, which would be helpful for writing code for tests. :)

Trying to use it with CMSSignedDataGenerator / JcaSignerInfoGeneratorBuilder still gives you:

```
org.bouncycastle.operator.OperatorCreationException: exception on setup: java.security.NoSuchAlgorithmException: no such algorithm: 1.3.14.3.2.29 for provider BC
```

Is it possible in addition for the changes you already made for you to pull in the [`digestOidMap` changes to `MessageDigestUtils.java`](https://github.com/bcgit/bc-java/pull/1500/files#diff-3e0cc724f70ca280ca9025255b777080badb2471fa397b9c72341be0fc523e22)? That allows the generators to resolve the alias correctly.

> Is it possible to provide a sample of one of these messages?

Here's an example of a random CMS message using this:

```pem
-----BEGIN PKCS7-----
MIAGCSqGSIb3DQEHAqCAMIACAQExCTAHBgUrDgMCHTCABgkqhkiG9w0BBwGggCSA
BBNNeSBzcGVjaWFsIG1lc3NhZ2UyAAAAAAAAoIAwggSZMIIDTaADAgECAhQAoX1E
JJLxftRRUu06RXzDwMewAzBBBgkqhkiG9w0BAQowNKAPMA0GCWCGSAFlAwQCAQUA
oRwwGgYJKoZIhvcNAQEIMA0GCWCGSAFlAwQCAQUAogMCASAwSDELMAkGA1UEBhMC
VVMxEzARBgNVBAoTCkdvb2dsZSBMTEMxJDAiBgNVBAMTG0dFTS1XaW5kb3dzLXVz
LXdlc3QxLUdZSm8yNTAeFw0yMzEwMTcyMTU2MjlaFw0yNDAxMjUyMTU2MjhaMBQx
EjAQBgNVBAMMCWRldmljZV9pZDCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoC
ggEBALGMmPlHHlrXysf43O6cKbDta0zmo5bWwCzir78rkfyUsSdBDC+GLRhq2bxr
AW8nYn6wDy0a2r2QMoj9SS4my38o2Urtu+XtsQHu0NG+hKxm7KLSRncpvqcfFc6A
FQ8eYbbkYRVRJm5r7jsYWE9IB7Kv46rUWofieuYHdDyPvKU3unmxlegKFDWcFVmo
cx4BiMGZ5zv+BtLfHxCRTf5k9r1y13gmTnwZzI+iwc9ZixTHJD6OejvvEcbjqPUm
q9OhBrx/eUZqDCjHoBcjtYuj24RAhZu55/GP0xQedvZfasEA70XLD5wkF1oM/vQl
Cu9STP0en4y75YMefDjRmygpVdcCAwEAAaOCAUUwggFBMA4GA1UdDwEB/wQEAwID
uDATBgNVHSUEDDAKBggrBgEFBQcDAjAMBgNVHRMBAf8EAjAAMB0GA1UdDgQWBBQ7
eoXi4IwFUvD++kLjKcWIbZM+tDAfBgNVHSMEGDAWgBTs9EezkU1W31lKwtOxtGEI
RaJSuTCBjQYIKwYBBQUHAQEEgYAwfjB8BggrBgEFBQcwAoZwaHR0cDovL3ByaXZh
dGVjYS1jb250ZW50LTY1MGE2MmE2LTAwMDAtMmU3Mi1hZGY1LWQ0ZjU0N2ZhY2Fh
MC5zdG9yYWdlLmdvb2dsZWFwaXMuY29tLzZiYjBlZmIwZjk5MmZmNmZkYTFkL2Nh
LmNydDA8BgNVHREENTAzhjFkYXRhOkNnbGtaWFpwWTJWZmFXUWdxK09QeVBvU09n
WUlnT0xKckFaQWdhK05BZz09MEEGCSqGSIb3DQEBCjA0oA8wDQYJYIZIAWUDBAIB
BQChHDAaBgkqhkiG9w0BAQgwDQYJYIZIAWUDBAIBBQCiAwIBIAOCAQEAroRz0SiP
2IMkwcZ0ffz/d4Z9RE9x7lDve2hsmwKqYnKzYmiueVognNGfI3zp613+NJN73Au4
qeCHUMmq+rik2m2qMy/1PqzUHthqpGONXfckw18+XpTqmSsBm+wGBvfjFAFi1e/Q
x64ABjLyGfws9/fiZgOkCc/ewDo+BI2S8vaYMRTVhKtfLZQODXTLQlalFX2JM0B6
KYOfOmfmNM/W0FQLOetbprSqCsA2Di2e3/obnh0N+ft2q3Ayvh9dEPNOnp0W1hAD
VAHhoq4HQKCsaJuc93pjIDgoFnFg5jBeKq2ZUDJc32N21iEPK2ylrUVnUVZx87wK
laDjdqbmrN0R2gAAMYICDjCCAgoCAQEwYDBIMQswCQYDVQQGEwJVUzETMBEGA1UE
ChMKR29vZ2xlIExMQzEkMCIGA1UEAxMbR0VNLVdpbmRvd3MtdXMtd2VzdDEtR1lK
bzI1AhQAoX1EJJLxftRRUu06RXzDwMewAzAHBgUrDgMCHaCBhjAYBgkqhkiG9w0B
CQMxCwYJKoZIhvcNAQcBMBwGCSqGSIb3DQEJBTEPFw0yMzEwMjMwMzE0NDRaMCMG
CSqGSIb3DQEJBDEWBBQ6a3vz6YBI8hq0TwAbAhqRH2Ry0TAnBgkqhkiG9w0BCTQx
GjAYMAcGBSsOAwIdoQ0GCSqGSIb3DQEBAQUAMA0GCSqGSIb3DQEBAQUABIIBAJc+
27Uik2JfE5xloR5Ry2FZsslXRrrO7PdhlXvwgPpgicma3AYvD9os7fk6nTieEaVB
fWkesz7fFOthP/UG5KvpsyE1hmXMqUpCPQfFE1XfzaSgBeMTRWtPqzagVojJw+3i
LrCy0Kgw+wKBwmyVkd0zHLQEF5aUA+ovCTNGje11qNpdCwUKfal7uz8TD4V0hWwJ
zAYME3itle90ztUh3NcEYQxJOQNZRGhkTMsXHKNPDCGS+erpU8BERkU+FOCH9+wS
CiwhMCLDeeEBOdxWZHVbIiFnnRTQqyIDGAOSSIUmjE/pMPKpPvumkCGq2r9GxPV9
YlpnThaYbDCnWg8tbWYAAAAAAAA=
-----END PKCS7-----
```

dghgit:
Thanks for the sample. With hindsight, I've got mixed feelings about the fact it works, I note it's using signed attributes, I guess since you've generated it locally - but for that particular use case it does mean we're weakening an already weak algorithm further.

With further changes, I hate break bad news, but the generator is resolving the alias correctly - producing messages like this should not be encouraged (I'll leave the idea of using SHA-1 in general now for another discussion...).

In the (hopefully) unlikely event you're actually supposed to send broken messages to who ever was responsible for this in the first place, one interesting difference between the generated sample and the message the dump was provided of early, is the original message used a raw signature and no signing attributes. You can actually create one of these using

JcaJcaSimpleSignerInfoGeneratorBuilder.setDirectSignature()

and then doing (for the simplest case):

        SignedData sd = SignedData.getInstance(ContentInfo.getInstance(s.toASN1Structure()).getContent());
        SignerInfo sInfo = SignerInfo.getInstance(sd.getSignerInfos().getObjectAt(0));
        sInfo = new SignerInfo(sInfo.getSID(), new AlgorithmIdentifier(OIWObjectIdentifiers.sha1WithRSA, DERNull.INSTANCE), sInfo.getAuthenticatedAttributes(), sInfo.getDigestEncryptionAlgorithm(), sInfo.getEncryptedDigest(), sInfo.getUnauthenticatedAttributes());

        sd = new SignedData(new DERSet(new AlgorithmIdentifier(OIWObjectIdentifiers.sha1WithRSA, DERNull.INSTANCE)),   sd.getEncapContentInfo(), sd.getCertificates(), sd.getCRLs(), new DERSet(sInfo));
   
        s = new CMSSignedData(new ContentInfo(PKCSObjectIdentifiers.signedData, sd));

One further question, are these message types always direct signatures? If possible, I would rather go back to failing for signatures using signed attributes, people having to rely on SHA-1 based signatures already have enough issues without us further compromising the security of the signature.

kevinhwang:
Ah you're right, I generated it locally with signed attributes. I wanted to avoid posting a real example of messages we're seeing because of PII.

Using direct signature works:

```kotlin
val cert: X509Certificate = ...
val privateKey: PrivateKey = ...

val contentSigner: ContentSigner =
  JcaContentSignerBuilder("SHA1WithRSAEncryption")
    .setProvider(BouncyCastlProvider())
    .build(privateKey)
val content: CMSTypedData = CMSProcessableByteArray("My special message2".toByteArray())
val signedData: CMSSignedData =
  CMSSignedDataGenerator()
    .apply {
      addSignerInfoGenerator(
        JcaSignerInfoGeneratorBuilder(
          JcaDigestCalculatorProviderBuilder().setProvider(BouncyCastlProvider()).build()
        )
          .setDirectSignature(true)
          .setContentDigest(AlgorithmIdentifier(OIWObjectIdentifiers.sha1WithRSA))
          .build(contentSigner, cert)
      )
      addCertificates(JcaCertStore(setOf<X509Certificate>(cert)))
    }
    .generate(content, true)
```

I updated my example in my comment above to be a CMS message using direct signature w/ the sha1WithRsa digest algo alias.

Issue #1508, opened 2023-10-16
bc-java 1.75 fails to parse (admittedly invalid) certificates that used to be parseable 


pjsg:
We use bc-java to parse certificates but 1.75 fails to parse some invalid certificates (which are mostly valid). Ideally bc would return more information about the failure rather than just throw an IllegalArgumentException.

Suggest that a subclass of IllegalArgumentException be thrown with more information about why the certificate was invalid. 

I can make this change.....

dghgit:
Yes, if you would like to submit a pull request, I'd be happy to look at it. Curious to know what the errors are, I'm guessing they may be about malformed integers?

pjsg:
I stuck up a WIP PR: #1511 which is the functional code that implements the change. I haven't added tests for the changes yet (the actual tests that I have been using are in our private repo which makes use of BC). If this approach seems reasonable, then I'll go ahead and do that.

Some of the issues are malformed integers, but there are others as well -- invalid version numbers etc.

winfriedgerlach:
@dghgit is this still being worked on?

dghgit:
We've taken a slightly different angle on this, have added an X509CertificateReviewer to the org.bouncycastle.cert package which will give a verbose (hopefully not too verbose) report on a certificate it is given, in combination with ideas in PR #1151. Now up on https://www.bouncycastle.org/betas

PR #1511, opened 2023-10-19
[WIP] Fixes Issue #1508 providing more information about certificate parsing issues

pjsg:
This PR adds an `IllegalArgumentWarningException` that is a subclass of `IllegalArgumentException` which is used to convey more information about certificate parsing errors. In particular, the Exception includes a list of all the errors that were encountered during the parsing, *and* includes (if possible) the parsed certificate object. 

This change is backwards compatible.

I haven't (yet) added tests or updated the documentation -- I'll do that if this approach seems worthwhile.


dghgit:
For RSAKeyParameter you can set:

"org.bouncycastle.rsa.allow_unsafe_mod" to true to avoid the validations

these are things that by rights would usually result in a CVE if they were accepted. If (for other reasons) it's necessary to accept such a key, you might as well be transparent about it.

If anything on the list fails though the key should be rejected outright, it's not actually safe to use (or put another way, it's probably not worth putting together a list of what's wrong with it, it's not going to help, the key is invalid and should be rejected, so defining special handling for the exception isn't going to help...).

I'm wondering looking at this if what you're really looking for is a reporting class, something like the CertPathReviewer, which can work through a certificate and explain what's wrong with it?

pjsg:
Yes -- I'm looking for a detailed reporting class and I didn't want to duplicate all the parsing code and thereby cause maintenance problems (as they would, inevitably) get out of sync. 

Do you have a suggestion on how to achieve this goal?

dghgit:
Thanks for the submission. We've taken a slightly different angle on this, have added an X509CertificateReviewer to the org.bouncycastle.cert package which will give a verbose (hopefully not too verbose) report on a certificate it is given with it's own checks a list based check on extensions and the TBS certificate. This should give space for further additions to pick up other things as well as avoiding any possible issues with classes relying on fail first/fast. We hope it will be regarded as merged with revision. Now up on https://www.bouncycastle.org/betas

PR #1517, opened 2023-10-25
Tls injection mechanism

sergejskozlovics:
I have developed a TLS injection mechanism. It can be used to inject PQC algorithms (KEMs and signature algorithms) into TLS 1.3. The injection mechanism is located in the org.bouncycastle.tls.injection package, where the InjectionPoint class has been defined. Then, there are many places where InjectionPoint is consulted for injected KEMs and signature algorithms, which are added to the corresponding lists of supported TLS algorithms.


peterdettman:
Hi Sergejs, thanks for the PR. It will be a while before a full review can be done, but on a first pass I have seen some basic issues:

- core project cannot depend on tls project (only the reverse). So I guess the basic injection concept would have to live in core and be used/built-upon by tls.
- tls project supports still minimum java version 5 (actually we still build also for jdk 1.4 with org.bouncycastle.tls classes only - no JSSE). I see at least some usage of Java 8 stream API, which should be replaced, and perhaps there are others. There are src/main/jdk??? folders to provide multi-release override classes if needed.
- There's some incidental reformatting that should be removed from the patch; I see various imports collapsed to ".*", braces moved, some switch case indentation, and there's probably more.

We are certainly interested in getting PQC algorithms into TLS, but our intention would be to proceed by publishing algorithms in the BC provider as usual, then having TLS add the registrar entries from the relevant standards and support them when it can find the algorithm via the provider. Perhaps you could comment a little about what you have in mind in terms of the basic "injection" idea here, particularly in comparison to the existing provider-based mechanism.

PR #1564, opened 2024-01-12
Implement Memoable Interface

seungyeonpark:
I noticed that among the classes implementing the Memoable interface, `CertPathValidationContext` and `CertificatePoliciesValidation` have not implemented the interface's methods appropriately.

I have updated the `copy` and `reset` methods so that these two classes can properly implement the Memoable interface.

I would appreciate it if you could review the proposed code. If there are any areas that need modification or if you have any additional suggestions, I would be grateful for your feedback.


dghgit:

PR #1575, opened 2024-01-27
Sanitize boolean conversions in signature subpackets.

vanitasvitae:
Currently, a `PrimaryUserID` packet whose boolean value is encoded as an invalid, but non-zero value (e.g. `2`) is interpreted as `true`.
I believe, BC should detect this invalid value as the sign that something went wrong and instead throw an exception.

This PR introduces the `Bytes` utility class in the core module, which contains two methods for converting between byte arrays and booleans.
Especially the `fromByteArray()` is more strict in that it does not allow arrays with a length other than 1 and also does not allow byte values other than 0 and 1.

### DIFF
--- a/core/src/main/java/org/bouncycastle/util/Booleans.java
+++ b/core/src/main/java/org/bouncycastle/util/Booleans.java
@@ -1,0 +1,44 @@
+package org.bouncycastle.util;
+
+public class Booleans {
+
+    /**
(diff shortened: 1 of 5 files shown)


vanitasvitae:
The same type of sanitization can also be applied to `Revocable` and `Exportable`.
I'll add separate commits for these.

dghgit:
Thanks for the patch. Merged!

PR #1596, opened 2024-03-06
Explicit targets for build-util and build-pkix

brownp2k:
Add explicit ant targets for build-util and build-pkix.

I found myself not needing to build the entire bouncycastle product but there was no way to avoid building everything, and further, a normal build (ant -f ant/jdk18+.xml build) fails due to not finding ASN1ObjectIdentifier, but that is a separate issue.

Adding these two explicit targets allows for more fine-grained control of building subprojects.


dghgit:
Thanks for the patch. Merged!

PR #1615, opened 2024-04-05
added name resolving for jurisdiction{C,ST,L}. These are used in EV c…

mtgag:
In EV certificates there are three attributes that are used in the subjectDN and are not resolved as a string for an X500Name. This PR adds them.


dghgit:
Thanks for the patch. Merged.

PR #1636, opened 2024-04-24
Add FingerprintUtil helper for deriving key-ids from fingerprints

vanitasvitae:
Separated out from #1634


dghgit:
Merged. Thanks!

PR #1637, opened 2024-04-24
Add HashUtils helper for looking up salt sizes for digest algorithms

vanitasvitae:
Separated out from #1634


dghgit:
Merged. Thanks!

PR #1642, opened 2024-04-29
Allow roundtripping OpenPGP packet format

vanitasvitae:
The OpenPGP protocol defines different formats to encode packet lengths.
BC supports reading and writing both packet format types, though it is not always clear, how to chose, which format to use.

This PR improves the status quo by defining an enum `PacketFormat` consisting of 3 options:
* `LEGACY`: Always encode packets using the old legacy packet format
* `CURRENT`: Always encode packets using the current (new) packet format
* `ROUNDTRIP`: Encode packets that were read from an `InputStream` the same way they were originally encoded.

Further, packets now contain a new flag remembering the packet format they were originally encoded in.
This, and the fact that there is a new `BCPGOutputStream` constructor allowing the user to chose which format encoding strategy to use allows to round-trip packet streams without changing the packet format.

`getEncoded()` now round-trips the original packet format, but the user can decide, in which format they want to retrieve the packet encoding by calling e.g. `getEncoded(PacketFormat.LEGACY)`.

The changes were introduced in a backwards compatible way and the tests still run successful.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/AEADEncDataPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/AEADEncDataPacket.java
@@ -23,9 +23,16 @@ public class AEADEncDataPacket
     private final byte[] iv;
 
     public AEADEncDataPacket(BCPGInputStream in)
+            throws IOException
+    {
+        this(in, false);
+    }
+
+    public AEADEncDataPacket(BCPGInputStream in,
+                             boolean newPacketFormat)
         throws IOException
     {
-        super(in, AEAD_ENC_DATA);
+        super(in, AEAD_ENC_DATA, newPacketFormat);
 
         version = (byte)in.read();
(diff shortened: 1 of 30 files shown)


dghgit:
I'm seeing a failure in OnePassSignaturePacketTest, would you take a look? Getting:

Expected: 
00000000  90 0d 03 00 08 01 00 00  00 00 00 00 00 7b 00 

Got: 
00000000  c4 0d 03 00 08 01 00 00  00 00 00 00 00 7b 00 

Looks like something may not be fully round-tripping.

Thanks.

vanitasvitae:
Ah, probably because you locally applied this PR over https://github.com/bcgit/bc-java/pull/1634
I did not yet update this PR based on that patch :D

Will update :)

vanitasvitae:
Rebased and fixed.

dghgit:
Thanks, now passes. Merged!

PR #1647, opened 2024-05-02
fix: allows developers to provide a block pool to argon2

### LINKED ISSUE #1646: Better memory management with argon2
The Keycloak project is needing to tweak garbage collection settings to deal the amount of garbage created by the Argon2BytesGenerator.

It would be good if there were a way to limit the amount of garbage created.

shawkins:
closes: #1646

Provides a way for developers to reuse memory for argon2. Assumes that the implementation will call allocate / deallocate in pairs - no checking is done for multiple deallocation. 

If you are open to this approach, what kind of tests would you like to see for this?

cc @ahus1 @vmuzikar


shawkins:
@jill-kleiber @dghgit Rebased this. Please let me know if there is any interest in this, or if it should just be abandoned.

jill-kleiber:
We're currently in the process of publishing our PQC work. One that's wrapped up I'll have a look at this.

dghgit:
Thanks for the patch, this has been added with a minor revision and some tests (moved the block pool definitions). Now in https://www.bouncycastle.org/betas

Apologies for the delay.

PR #1658, opened 2024-05-10
PGPKeyConverter.implGetKdfParameters(): Add support for X448

vanitasvitae:
See https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html\#name-algorithm-specific-fields-for-x for reference.


vanitasvitae:
Closed in favor of #1663

PR #1664, opened 2024-05-15
Abstract core KEM functionality out of DHKEM

jmcrawford45:
Hi, I'd like to be able to use HPKE with alternative KEMs, but it's currently not usable because there's no public facing KEM  abstraction and HPKE currently hard codes the initial set from RFC9180. To accomplish this, I've pulled the public functions from DHKEM into the KEM abstract class, modified HPKE to use a KEM instead, and added a new constructor that doesn't require an RFC9180 KEM.


jmcrawford45:
Hi @roy-basmacier,

I'm a bit confused as to why the pull request was closed without any comment or merge. Could you please provide some insight or feedback on this? I want to make sure I'm following the correct procedures and address any issues that might have led to this decision.

Thank you!

PR #1663, opened 2024-05-14
Rewrite PGPKeyConverter classes to fix support for X448,Ed448,X25519,…

vanitasvitae:
…Ed25519 keys

Obsoletes #1658 

This patch
* Adds support for X448 in `PGPKeyConverter.implGetKdfParameters()`
* Adds support for legacy X448 in `BcPGPKeyConverter`
* Adds support for legacy Ed448 in `BcPGPKeyConverter.getPrivateBCPGKey()`
* Rewrites `BcPGPKeyConverter.getPublicBCPGKey()` to detect key types based on algorithm ID
  * Fixes proper return types
    * legacy Ed448: `Ed448PublicBCPGKey` -> `EdDSAPublicBCPGKey`
    * legacy X448: `X448PublicBCPGKey` -> `ECDHPublicBCPGKey`
* `JcaPGPKeyConverter.getPGPPublicKey()`: 
  * Add detection of legacy X25519 using OID 1.3.101.110
  * Add support for legacy X448
  * Add support for legacy Ed448
* `JcaPGPKeyConverter.getPublicKey()`:
  * Add support for legacy X448
  * Add support for legacy Ed448
* Rewrite `JcaPGPKeyConverter.getPublicBCPGKey()` to detect key types based on algorithm ID
  * Add support for converting `XDHPublicKeyImpl` keys
  * Add support for legacy Ed448
  * Add support for legacy X448
* Fixes X25519, X448 keys (using the new dedicated `PublicKeyAlgorithmTags`) to use native encoding for the private key material
* Fixes generation of legacy Ed448 keys (using `PublicKeyAlgorithmTags.EDDSA_LEGACY` together with an Ed448 key (`BC15EdDSAPublicKey` passed into `JcaPGPKeyConverter`) caused `PGPException` (unknown key class))
* Fixes conversion of legacy Ed25519 keys (if `setProvider()` is not called on `JcaPGPKeyConverter`)
  * `BcPGPKeyPair` of Ed25519 key + ECDH algorithm tag cannot be converted to `JcaPGPKeyPair`
    * `EdDSAPublicKeyImpl` is not recognized in `JcaPGPKeyConverter.getPublicBCPGKey()` -> `PGPException("unknown key class")`
* Fixes generation of legacy X448 keys (`JcaPGPKeyPair`, `PublicKeyAlgorithmTags.ECDH` together with an X448 key)
  * `BC11XDHPublicKey` is not recognized in `JcaPGPKeyConverter.getPublicBCPGKey()`
[...]

PR #1671, opened 2024-05-21
Fix PGP EC Key Conversion

vanitasvitae:
This PR is based on #1663 and fixes JcaPGPKeyConverter conversions for Elliptic Curve keys (Brainpool, NIST etc.).

The test from aeb001d4b9556502fc2205babef9d23375cacf87 shows that conversion between Jca and BC keys fails for some elliptic curve based keys, because some keys report the wrong OID (`X9ObjectIdentifiers.id_ecPublicKey`).
The patch from 87bbd5db7218e69cb397e47d9183a2a7f8db7f6f drills down on the key to detect the proper OID by iterating over all known curves (a bit inefficient, is there a cleaner way?).

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPKeyConverter.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPKeyConverter.java
@@ -557,10 +557,10 @@ public class JcaPGPKeyConverter
                 ASN1ObjectIdentifier curveOid;
                 curveOid = ASN1ObjectIdentifier.getInstance(enc);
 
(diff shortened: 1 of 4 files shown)


vanitasvitae:
Some observations:
JcaPGPKeyConverter still fails to convert EC keys over prime256v1 and P-256 if no BouncyCastleProvider is set.

dghgit:
Thanks. Merged with revision - you'll find the changes in JcaPGPKeyConverter, I've generalized it to deal with non-BC keys and also added use of the X962Parameters type which is the ASN.1 construct you'll find in a key of type id_ecPublicKey. It would be worth having a look as it explains what's going on internally in the actual Java providers (well at least the ones I have seen to date).

PR #1675, opened 2024-05-23
Fix EDDSA_LEGACY signing/verification with Ed448 keys

vanitasvitae:
When creating/verifying signatures made using `PublicKeyAlgorithmTags.EDDSA_LEGACY`, BC creates Signer object instances of Ed25519. While this is in line with the rfcs (EdDSALegacy is only specified for Ed25519), BC also supports key generation etc. using Ed448.
(Do we want to retain support for LEGACY_EDDSA + Ed448?)

(Update: LEGACY_EDDSA + Ed448 is defined in LibrePGP, see #1677 )

As a result, when generating/verifying signatures made using Ed448+EDDSA_LEGACY, BC tries to use Ed25519 signers, which fails.
Further, PGPSignature parses EDDSA_LEGACY signatures only as Ed25519, which means Ed448 signature result in OOB exceptions.

This PR changes the signer/verifier methods to detect, which curve is actually used when a EDDSA_LEGACY key is provided and to return an appropriate signer object accordingly.
The patch also fixes signature parsing for EDDSA_LEGACY Ed448 signatures by inspecting the length of the encoded signature values.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
@@ -19,6 +19,7 @@ import org.bouncycastle.bcpg.SignaturePacket;
 import org.bouncycastle.bcpg.SignatureSubpacket;
 import org.bouncycastle.bcpg.TrustPacket;
 import org.bouncycastle.math.ec.rfc8032.Ed25519;
+import org.bouncycastle.math.ec.rfc8032.Ed448;
 import org.bouncycastle.openpgp.operator.PGPContentVerifier;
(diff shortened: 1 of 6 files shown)


dghgit:
Usually the best way to deal with this is to support verification, but disable generation, so for now at least the best thing to do would be to revert the ability to use an Ed448 key in this context, but allow the signatures generated that way to be verified.

vanitasvitae:
Apparently, EDDSA_LEGACY+Ed448 is used by LibrePGP, so its debatable, whether it makes sense to deliberately revert signing support, now that its already implemented. I'd leave this decision up to you :)

dghgit:
Far be it from me to argue with the market on this one! Lets leave it as it be and hope for the best.

PR #1687, opened 2024-05-28
Prevent PGP Signature Type 0xFF

vanitasvitae:
The Crypto-Refresh specification [states](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-reserved-type-id-0xff):
> An implementation MUST NOT create any signature with this type \[0xFF\], and MUST NOT validate any signature made with this type.

This PR prevents the generation and verification of signatures with type 0xff.


dghgit:
Merged. Thanks!

PR #1693, opened 2024-06-02
Test and fix conversion of EdDSA keys with leading zero

### LINKED ISSUE #1690: Converting Ed25519, Ed448 keys fails if encoding has leading 0
Hey!
I noticed, that generating Ed448 keys is not totally reliable and fails in 0.5% of the time.

Apparently, in some cases, [in this block](https://github.com/bcgit/bc-java/blob/main/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/edec/KeyFactorySpi.java#L179-L191), the type is identified as Ed448 by looking at `enc[8]`, but the prefix array does not match the header bytes of the `enc` array, causing [this check](https://github.com/bcgit/bc-java/blob/main/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/edec/BCEdDSAPublicKey.java#L40) to fail.

An example:
The conversion fails for a key with the following encoding 
```
rawData = 3042300506032b6571033900cc41697d47dd92200f746bded5cd456274cd5a879ed8e2ba3c46c943b532316b115a5173ba5fae196cdce5bcc716295f24b8acdfb7ea6280
prefix =  3043300506032b6571033a00
```
Note the mismatch in octets with index 1 and 10.

Contrary, a valid key:
```
rawData = 3043300506032b6571033a004ead091781d6ae7d7ad606aa6846e3e03d85b3b12d777e69e791d648cd7d2a16195e1f50a06f6d3600baa4460df216e83e5e256c443ee53880
```

vanitasvitae:
This PR contains a test and a patch for conversion of Ed25519, Ed448 keys which have a leading 0 in their encoding.
Fixes #1690

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPKeyConverter.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPKeyConverter.java
@@ -367,22 +367,13 @@ public class JcaPGPKeyConverter
             // Modern Ed25519 (1.3.6.1.4.1.11591.15.1 & 1.3.101.112)
             case PublicKeyAlgorithmTags.Ed25519:
             {
-                BCPGKey key = publicPk.getKey();
-                if (key instanceof Ed25519PublicBCPGKey)
-                {
-                    return implGetPublicKeyX509(BigIntegers.asUnsignedByteArray(new BigInteger(1, publicPk.getKey().getEncoded())),
+                return implGetPublicKeyX509(publicPk.getKey().getEncoded(),
                         0, EdECObjectIdentifiers.id_Ed25519, "EdDSA");
-                }
-                else
-                {
-                    return implGetPublicKeyX509(BigIntegers.asUnsignedByteArray(((EdDSAPublicBCPGKey)publicPk.getKey()).getEncodedPoint()),
-                        0, EdECObjectIdentifiers.id_Ed25519, "EdDSA");
-                }
             }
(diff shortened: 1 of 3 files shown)


vanitasvitae:
I added a test for an Ed25519 key with a leading 0 as well.

pg/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPKeyConverter.java
```diff
@@ -367,22 +367,13 @@ else if (ecdhK.getCurveOID().equals(EdECObjectIdentifiers.id_X448))
...
             {
-                BCPGKey key = publicPk.getKey();
-                if (key instanceof Ed25519PublicBCPGKey)
-                {
-                    return implGetPublicKeyX509(BigIntegers.asUnsignedByteArray(new BigInteger(1, publicPk.getKey().getEncoded())),
-                        0, EdECObjectIdentifiers.id_Ed25519, "EdDSA");
-                }
-                else
-                {
-                    return implGetPublicKeyX509(BigIntegers.asUnsignedByteArray(((EdDSAPublicBCPGKey)publicPk.getKey()).getEncodedPoint()),
```
vanitasvitae, line 378: Since `PublicKeyAlgorithmTags.Ed25519` is only used with `Ed25519PublicBCPGKey`, I was able to remove the if-else statement.

dghgit:
This makes sense. Merged! Thanks.

PR #1694, opened 2024-06-03
Fix .github/workflows/codeql.yml

vanitasvitae:
The file had the `- run:` statement, which spanned across multiple lines, not completely commented out, resulting in syntax errors, which trigger emails being sent with each pushed commit.


vanitasvitae:
In main as 4472df4a03b5078ef74ebc601758cbe107386144

PR #1696, opened 2024-06-04
Add support for unknown PGP key types

vanitasvitae:
This PR adds support for OpenPGP (v6) keys with unknown public key algorithms.

The v6 test key
```
-----BEGIN PGP PUBLIC KEY BLOCK-----

xhQGZl7182MAAAAKAAECAwQFBgcICQ==
=anvn
-----END PGP PUBLIC KEY BLOCK-----
```
contains a key with unknown algorithm `99` and consists of the octets `0-9`:
```
Public-Key Packet, new CTB, 2 header bytes + 20 bytes
    Version: 6
    Creation time: 2024-06-04 11:09:39 UTC
    Pk algo: Unknown algo 99
    Fingerprint: ACD6CEC3A014079F78DA9A60B98F9E3D4E7321834E1EBEC68A1C3701F882D5A3
    KeyID: ACD6CEC3A014079F
  
    00000000  c6                                                 CTB
    00000001     14                                              length
    00000002        06                                           version
    00000003           66 5e f5 f3                               creation_time
    00000007                       63                            pk_algo
    00000008                           00 00 00 0a               public_len
    0000000c                                       00 01 02 03   rest
    00000010  04 05 06 07 08 09
```

Handling unknown keys gracefully is possible with v6 keys, since the key encoding length is known.

TODO: Figure out, if we can also support unknown v4/v5 keys.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyPacket.java
@@ -60,10 +60,10 @@ public class PublicKeyPacket
         }
 
         algorithm = (byte)in.read();
+        long keyOctets = 0;
         if (version == VERSION_6)
         {
-            // TODO: Use keyOctets to be able to parse unknown keys
-            long keyOctets = ((long)in.read() << 24) | ((long)in.read() << 16) | ((long)in.read() << 8) | in.read();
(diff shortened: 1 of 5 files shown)


vanitasvitae:
We can only gracefully parse v5, v6 keys, since secret key packets simply append the secret key material after the public key material, so when parsing a v4 secret key, we don't know, when to stop parsing the public key material.
V5 and v6 keys include an additional field encoding the public key length to enable graceful parsing (as implemented here).

PR #1701, opened 2024-06-11
S2K: reuse type constants

vanitasvitae:


vanitasvitae:
Changes are part of main already.

PR #1702, opened 2024-06-11
Implement secret key protection using AEAD

vanitasvitae:
This PR is based on - and blocked by - #1699

Here, I added two new classes, BcAEADSecretKeyEncryptorBuilder and JcaAEADSecretKeyEncryptorBuilder, which can both be used to AEAD-encrypt PGP secret keys.

First I tried to implement AEAD protection into the existing PBESecretKeyEncryptorBuilder classes, but that turned out to result in ugly code.
Let me know what you think of my approach.


vanitasvitae:
Blocked by https://github.com/bcgit/bc-java/pull/1695 because we need to access the new versioned `BcPGPKeyPair`, `JcaPGPKeyPair` constructors.

PR #1704, opened 2024-06-12
Ordered attribute table

ebourg:
I'd like to suggest a modification to the AttributeTable class such that the insertion order of the attributes is preserved. The current implementation uses a Hashtable and doesn't guarantee any order. This PR replaces the Hashtable with a LinkedHashMap.


ebourg:
Actually that's pointless since the DERSet is sorted afterward. Sorry for the noise. The first three commits might be still worth picking though.

PR #1706, opened 2024-06-12
Test PGPSignatures with LegacyEd25519 keys with short MPIs

vanitasvitae:
With legacy ed25519 we need to detect the curve (curve25519 / curve448) based on the length of the encoding (see https://github.com/bcgit/bc-java/pull/1675, most notably https://github.com/bcgit/bc-java/pull/1675/commits/105dbef01d688a963f98ba2815f1343774c42fae).

However, I noticed, that some ed25519 signatures MPIs are shorter than the expected 2*32 octets (leading 0s), resulting in the code misinterpreting the signature as an ed448 signature. This in turn causes signature verification failures for good signatures.

This patch catches this edge case by allowing short MPIs.

### DIFF
--- a/pg/src/test/java/org/bouncycastle/openpgp/test/PGPSignatureTest.java
+++ b/pg/src/test/java/org/bouncycastle/openpgp/test/PGPSignatureTest.java
@@ -11,13 +11,22 @@ import java.security.SignatureException;
 import java.util.Date;
 import java.util.Iterator;
 
-import org.bouncycastle.bcpg.*;
+import org.bouncycastle.bcpg.ArmoredInputStream;
+import org.bouncycastle.bcpg.BCPGInputStream;
+import org.bouncycastle.bcpg.CompressionAlgorithmTags;
+import org.bouncycastle.bcpg.HashAlgorithmTags;
+import org.bouncycastle.bcpg.PublicKeyAlgorithmTags;
+import org.bouncycastle.bcpg.SignatureSubpacket;
+import org.bouncycastle.bcpg.SignatureSubpacketInputStream;
+import org.bouncycastle.bcpg.SignatureSubpacketTags;
+import org.bouncycastle.bcpg.SymmetricKeyAlgorithmTags;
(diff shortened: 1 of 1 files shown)


pg/src/main/java/org/bouncycastle/openpgp/PGPSignature.java
```diff
@@ -455,7 +455,7 @@ else if (getKeyAlgorithm() == PublicKeyAlgorithmTags.EDDSA_LEGACY)
             {
                 byte[] a = BigIntegers.asUnsignedByteArray(sigValues[0].getValue());
                 byte[] b = BigIntegers.asUnsignedByteArray(sigValues[1].getValue());
-                if (a.length + b.length == Ed25519.SIGNATURE_SIZE)
```
vanitasvitae, line 458: Here, `a.length` or `b.length` could be `31` instead of the the expected `32`.
Since Ed25519 signatures are shorter than Ed448 signatures, this comparison should hold.

vanitasvitae:
The test vector ([see as PGP packet here](https://dump.sequoia-pgp.org/?data=-----BEGIN%20PGP%20SIGNATURE-----%0D%0AVersion%3A%20PGPainless%0D%0A%0D%0AiHQEARYKACcFAmZqLUAJEFrFuD8aWtaHFiEEIpz8hf4MouNxiwIsWsW4Pxpa1ocA%0D%0AAKFrAPd1TB0UsGiuXmgWw3Y2dWmxrphFh%2Bjl7DzFS4EVSaSSAQDKIVnlllv32GVT%0D%0AhUSZlK6tFMzwXD8zM1uY0wXA8g71Dg%3D%3D%0D%0A%3D%2BfMz%0D%0A-----END%20PGP%20SIGNATURE-----&mpis=on)) demonstrates the issue by having a short 31-octet `r` value.

vanitasvitae:
Fix is already part of main, so I reduced the PR to only contain the test.

vanitasvitae:
In main as c51be492536977c46962bb0b80be178454e9f91c

PR #1710, opened 2024-06-17
Respond with TLS alert "bad_certificate" if certificate parsing fails

markus-jung-vivavis:
Improve the error handling in case a invalid (syntactically incorrect) certificate is received in a TLS handshake from the server. 

Previously, the exception thrown during ASN.1 parsing was not caught and therefore caused a TLS alert "internal_error". This PR handles the exception to respond with a RFC-compliant TLS alert "bad_certificate".


peterdettman:
The basic idea is good, but the patch is a bit simplistic:

- we should deal also with processing of the client certificate
- shouldn't swallow existing TlsFatalAlert (which is an IOException subclass)
- probably not all errors from Certificate.parse should be bad_certificate; I guess decode_error should be used for some cases and there may be others.

PR #1711, opened 2024-06-18
Document PublicKeyPacket members

vanitasvitae:
This PR adds javadoc documentation to all members of `PublicKeyPacket`.
The only non-documentation change is the introduction and use of the `LIBREPGP_5` version number constant.


vanitasvitae:
Already part of main.

PR #1736, opened 2024-07-09
FingerprintUtil: Add prettifyFingerprint method

vanitasvitae:
This PR adds a new method to `FingerprintUtil` which can be used to prettify OpenPGP fingerprints.

Fingerprints are getting hex encoded and split into chunks of 4 or 8 characters for easier comparison.


vanitasvitae:
I noticed, that checkstyle complains about the use of `Hex.toHexString()`. Can we ignore this warning or is there a better alternative to hex encode data to a string?

PR #1742, opened 2024-07-16
Document `OpenBSDBCrypt` password handling

Marcono1234:
Previously the behavior was not explicitly documented, and it might not have been obvious that passwords longer than 72 bytes will be truncated.

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/generators/OpenBSDBCrypt.java
+++ b/core/src/main/java/org/bouncycastle/crypto/generators/OpenBSDBCrypt.java
@@ -12,7 +12,10 @@ import org.bouncycastle.util.Strings;
  * Password hashing scheme BCrypt,
  * designed by Niels Provos and David Mazières, using the
  * String format and the Base64 encoding
- * of the reference implementation on OpenBSD
+ * of the reference implementation on OpenBSD.
+ * <p>
+ * Passwords are encoded using UTF-8. Encoded passwords longer than
+ * 72 bytes are truncated and all remaining bytes are ignored.
  */
 public class OpenBSDBCrypt
 {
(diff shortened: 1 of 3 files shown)


Marcono1234:
Maybe it would also be good to apply this to https://github.com/bcgit/bc-csharp, but I am not familiar with C#. So if possible, could you do that please?

core/src/main/java/org/bouncycastle/crypto/generators/OpenBSDBCrypt.java
```diff
@@ -12,7 +12,10 @@
  * Password hashing scheme BCrypt,
  * designed by Niels Provos and David Mazières, using the
  * String format and the Base64 encoding
- * of the reference implementation on OpenBSD
+ * of the reference implementation on OpenBSD.
+ * <p>
+ * Passwords are encoded using UTF-8. Encoded passwords longer than
```
Marcono1234, line 17: Regarding UTF-8: Do you think this wording here is misleading and it should say that this only applies when the password is provided as `char[]` (but not when provided as `byte[]`)? Or is it fine like this?

dghgit: I agree it's going to confuse some people. I've clarified the comment to say UTF-8 only applies for char[].

Marcono1234:
Thanks for merging @dghgit!

Regarding https://github.com/bcgit/bc-csharp, could you please adjust the documentation there as well (I am not familiar enough with C#)? Or would you prefer if I created a GitHub issue there so that this task it as least tracked?

dghgit:
I updated C# as well. You should see the update appear on github shortly.

PR #1743, opened 2024-07-17
Add LibrePGP PreferredEncryptionModes signature subpacket

vanitasvitae:
I decided to implement this in its own dedicated class to express that this is not an official OpenPGP signature subpacket.


vanitasvitae:
This PR now includes changes from #1464

PR #1744, opened 2024-07-17
Fix v4 skesk argon2 test vectors

vanitasvitae:
The `Argon2S2KTest` contained a test vector, where the wrong sample data was pasted.
This PR replaces the contents of the v4 SKESK with Argon2 + AES256 with the proper test vector (it used to contain the AES192 vector by accident).


vanitasvitae:
In main as 3c9c1faa960a57e60c72332b707d14b3140768c6

PR #1752, opened 2024-07-20
Introduce KeyIdentifier class

vanitasvitae:
For historic reasons, many places in the OpenPGP API and specification rely on 64-bit key-ids.
However, later versions of the standard migrate to use fingerprints (20 or 32 bytes) to reduce the risk of attacks based on the birthday paradox (key-id collisions).

In this PR I introduce the `KeyIdentifier` class, which can be created from both a key-id or a fingerprint, and which shall be used as a common stand-in for either key-id or fingerprints.
This enables a smooth transition from using key-ids to using fingerprints.

TODO:
* [ ] Implement tests
* [ ] integrate `KeyIdentifier` into existing API to transition away from pure key-ID uses


vanitasvitae:
In main as a7930a250edb83c625c9a0f5297456272e8cf25d

PR #1753, opened 2024-07-22
Update Poly.java

chchen-scholar:
Fix the bug


roy-basmacier:
This is not a bug, but the latest change NIST announced during their 5th PQC conference. More information can be found below

https://github.com/bcgit/bc-java/issues/1734#issuecomment-2214314034

PR #1756, opened 2024-07-24
Replace StringBuffer with StringBuilder

stratozero:
StringBuffer is a synchronized Appendable, useful when more than one publisher is appending text to the same Appendable, but a waste of resources when the object is created and consumed by the same method (e.g. various `toString()`).

Furthermore, in Java 21+ synchronization interferes with the pinning of virtual threads on the carrier thread, potentially slowing down the application if, while locking an object, a suspendable operation is performed (e.g. IO read / write or `Thread.sleep()`)


stratozero:
Am I missing something? Did I pull to the wrong branch? The PR is 5 months old and I haven´t received any notification about. I'm starting thinking that's the wrong way to contribute to BC java, is it?

peterdettman:
At the moment the patch is in an unreasonable state to merge (or even review). I am unsure if that was the case when it was originally submitted. It appears to at least need to be rebased, but possibly re-submitting in a new PR would be a better idea (but see below).

I believe the reason we have not previously updated to StringBuilder in this way is that it is only available from 1.5, whereas we still have builds for java versions all the way back to J2ME (which StringBuffer works fine for). In our build system this could mean maintaining separate (e.g. src/main/j2me) versions of files if we change the src/main/java one for the sole purpose of using StringBuilder.

Ordinarily StringBuffer wouldn't be involved in performance-critical parts of the library, so it's doubtful we would want to make the change except for files only used in 1.5+ builds.

stratozero:
Good point, thanks! I'll try to restart from scratch and edit only the points in which a simple string composition is applicable. In the meanwhile I'm closing the PR

PR #1775, opened 2024-08-14
fix: Fix counter increment error when using seekTo in SIC/CTR mode

han-jl:
When encrypting and decrypting in sequence, carry forward when the counter is **-1 to 0**, and carry forward when use the _seekTo_ is **127 to -128**. Because the counters calculated at the same progress are different, random read and write operations cannot be supported when encrypting and decrypting files.My English is poor, I hope you can understand.
### Reproducing and testing code

```java
    private final byte[] counter = new byte[]{0, 0};
    private byte[] counter1 = null;
    private byte[] counter2 = null;

    private void incrementCounterAt(int pos) {
        int i = counter2.length - pos;
        while (--i >= 0) {
            if (++counter2[i] != 0) {
                break;
            }
        }
    }

    private void incrementCounter() {
        int i = counter1.length;
        while (--i >= 0) {
            if (++counter1[i] != 0) {
                break;
            }
        }
    }

    private void incrementCounter(int offSet) {
        counter2 = counter.clone();
        byte old = counter2[counter2.length - 1];

        counter2[counter2.length - 1] += offSet;

        if (old != 0 && counter[counter.length - 1] < old)
        {
            incrementCounterAt(1);
        }
    }

    public void test() {
        int failNum = 0;
        for (int i = 0; i < 255; i++) {
            counter[1] += i;
            counter1 = counter.clone();
            for (int j = 1; j < 255; j++) {
                incrementCounter();
                incrementCounter(j);
                if (!Arrays.equals(counter1, counter2)) {
                    failNum++;
                }
            }
        }
        System.out.println("failNum=" + failNum);
    }
```
result: `failNum=32258`
Change the judgment condition to `if ((old & 0xff) + offSet > 255) `   
result: `failNum=0`

### Simulate the scene of file append

```java

public class Test {
    static byte[] fileBytes = new byte[0];
    static byte[] iv = new byte[]{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};

    public static void main(String[] args) {
        Random random = new Random();
        for (int i = 0; i < 255; i++) {
            iv[iv.length - 1] += i;

            String inStr = " 1234567890jl字符串";
            for (int j = 1000; j > 0; j--) {
                inStr += (char) ('a' + random.nextInt(26));
                verify(inStr);
                fileBytes = new byte[0];
            }
        }
    }

    private static void verify(String inStr) {
        int strLen = inStr.length();
        System.out.println("strLen " + strLen);
        byte[] bytes = inStr.getBytes(StandardCharsets.UTF_8);
        System.out.println("bytes len " + bytes.length);

        appendFile(bytes);
        appendFile(bytes);
        appendFile(bytes);

        byte[] out = new byte[fileBytes.length];
        newCipher().processBytes(fileBytes, 0, fileBytes.length, out, 0);
        String outStr = new String(out, StandardCharsets.UTF_8);

        if (!outStr.equals(inStr + inStr + inStr)) {
            System.out.println(inStr.length());
            System.out.println(outStr.length());
            System.out.println(Arrays.toString(iv));
            System.out.println(" outStr=" + outStr);
            throw new RuntimeException("fail");
        }
    }

    private static void appendFile(byte[] bytes) {
        SICBlockCipher cipher = newCipher();
        cipher.seekTo(fileBytes.length);
        byte[] out = new byte[bytes.length];
        cipher.processBytes(bytes, 0, bytes.length, out, 0);

        byte[] newFileBytes = Arrays.copyOf(fileBytes, fileBytes.length + out.length);
        System.arraycopy(out, 0, newFileBytes, fileBytes.length, out.length);
        fileBytes = newFileBytes;
    }

    private static SICBlockCipher newCipher() {
        SICBlockCipher sicBlockCipher = new SICBlockCipher(new AESEngine());
        byte[] key = "1234567890123456".getBytes();
        ParametersWithIV parametersWithIV = new ParametersWithIV(new KeyParameter(key), iv);
        sicBlockCipher.init(true, parametersWithIV);
        return sicBlockCipher;
    }
}

```

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/modes/SICBlockCipher.java
+++ b/core/src/main/java/org/bouncycastle/crypto/modes/SICBlockCipher.java
@@ -244,7 +244,7 @@ public class SICBlockCipher
 
         counter[counter.length - 1] += offSet;
 
-        if (old != 0 && counter[counter.length - 1] < old)
+        if ((old & 0xff) + offSet > 255)
         {
             incrementCounterAt(1);
         }


jill-kleiber:
Can you clarify what you are referring to by "carry forward when the counter is -1 to 0"? Thanks.

han-jl:
I don't know how to express it, but the number at the current position has reached the target, and the previous one should have changed. For example，the current counter is［0,0,0,0,0,0,0,-1］,next counter is ［0,0,0,0,0,0,1,0］,next counter is ［0,0,0,0,0,0,1,1］.When using seekTo or skip function,the previous digit have changed at 127 to -128.



---Original---
From: ***@***.***&gt;
Date: Fri, Aug 23, 2024 20:47 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [bcgit/bc-java] fix: Fix counter increment error when usingseekTo in SIC/CTR mode (PR #1775)




 
Can you clarify what you are referring to by "carry forward when the counter is -1 to 0"? Thanks.

peterdettman:
@jill-kleiber Basically the bytes of `counter` should be treated as unsigned values, but the existing code of `incrementCounter(int)` treats them as signed. So I think the patch is correct.

peterdettman:
Merged, thanks for the PR.

PR #1790, opened 2024-08-27
open GMSignatureSpi constructor to subclass

moonfruit:
I want to create a `GMSignatureSpi` subclass with customized `SM2Signer`. But `GMSignatureSpi`'s constrcutor is not opened to subclass. Please accept this to open it.


dghgit:
This one is a little odd, but within parameters for bc-java. Thanks, now merged!

PR #1793, opened 2024-08-28
Fix parsing of LibrePGP key-id

vanitasvitae:
Hey!
When rewriting the public key parser logic, I accidentally missed parsing out the key-id of LibrePGP (v5) keys.
This patch fixes this slip-up and adds a test.


vanitasvitae:
Closed in favor of #1809

PR #1794, opened 2024-08-28
Add auth enveloped recipients for KEK and KeyAgree

bukka:
Currently it is possible to use CMSAuthEnvelopedDataParser only with KeyTrans. This adds support for KEK and KeyAgree. The AEADInputDecryptor is separated to CMSInputAEADDecryptor as it is the same for all recipient types. A new test is added to properly test CMSAuthEnvelopedDataParser usage.


bukka:
@ligefeiBouncycastle would you be able to check this when you have time? It's an addition to your commit https://github.com/bcgit/bc-java/commit/3eb6ed55c3dedfb3b3280e59dd3450adb378d8b3

PR #1795, opened 2024-08-29
Fix LibrePGP SKESK5 + OCBED1 decryption

vanitasvitae:
While playing with some test vectors from the LibrePGP document, I noticed that decryption of message encrypted using a v5 SKESK (`SymmetricKeyEncSessionPacket`) + OCBED (`AEADEncDataPacket`) packets would fail decrypting.

It turns out that LibrePGP omits the additional HKDF key derivation step introduced by OpenPGP v6.
This PR adds the test vectors from the LibrePGP document and fixes decryption.


vanitasvitae:
Closed in favor of #1809

PR #1828, opened 2024-09-06
OpenPGP - Partial length encoding: Enforce minimum length of the first chunk

vanitasvitae:
The OpenPGP specification mandates that if partial body encoding is used, the first chunk must be at least 512 bytes long.
Currently, BC does not enforce this requirement (see https://tests.sequoia-pgp.org/#Tests_support_for_partial_body_encoding )

This PR adds a check for this requirement, which will throw an `IOException` for malformed packets.


vanitasvitae:
This patch currently breaks with a (broken) test vector from `BcPGPPBETest`.
I'm thinking of ways to make the chunk length check configurable to allow parsing of such broken messages.

PR #1832, opened 2024-09-10
Improve test coverage of PGP (v6) keys

vanitasvitae:
This PR changes the `PGPv6KeyTest` to more extensively cover parsing of different fields, like key-id, creation time, s2k etc.

Testing also showed, that `getBitStrength()` returned `0` for (dedicated X25519), X448, Ed25519, Ed448 keys.
It also returned `0` for unknown keys.

This PR fixes these issues by properly parsing the bit strength (as in key-size, not security-level).


vanitasvitae:
Oops, I noticed that during a previous rebase I got some unrelated commits included in the patch.
Fixed by rebasing once more on main.

PR #1836, opened 2024-09-14
Fix unpacking time (unsigned int) from octets for large values

vanitasvitae:
Time, as well as offsets (e.g. expiration) in OpenPGP are stored as seconds in unsigned Integers.

I noticed, that https://github.com/bcgit/bc-java/commit/8385a2ca13573435a64d6bd96769c97e2e9edee8 broke some tests in PGPainless that dealt with large expiration time intervals.

Turns out, the changed code accidentally converted the values to signed ints, which break for large (but legal) intervals.

The patch converts back to an unsigned value.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyPacket.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/PublicKeyPacket.java
@@ -134,7 +134,7 @@ public class PublicKeyPacket
             throw new UnsupportedPacketVersionException("Unsupported Public Key Packet version encountered: " + version);
         }
 
-        time = StreamUtil.read4OctetLength(in);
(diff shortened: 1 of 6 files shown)


vanitasvitae:
I wonder if more places in the code are affected by this, e.g. key creation time...

vanitasvitae:
`PublicKeyPacket` and `SignaturePacket` were also affected by this.

I also discovered a discrepancy between `StreamUtils.writeTime()` (expecting seconds) and `StreamUtils.readTime()` (returning millis).

I fixed this by changing `writeTime()` to expect millis and introducing `writeSeconds()` and `readSeconds()` dealing with seconds.

vanitasvitae:
Okay, now I wonder, if large values are handled properly in general (e.g. large lengths) :D

PR #1842, opened 2024-09-18
Ellipitic Curve implementation of JPAKE (Password-Authenticated Key Exchange by Juggling)

dawmit:
Took the current JPAKE implementation in Bouncy Castle using prime order groups and adjusted it to work with elliptic curves instead, specifically curves of short-weierstrass form. 

I have included and example use case of the ecjpake implementation, as well as test files files.


dawmit:
Apologies for the messy commits, the last 7 commites are just adjustments to comments. The first 3 commits are the important ones.

PR #1864, opened 2024-10-10
System property for enabling GCM ciphers for FIPS

crlorentzen:
This property allows enabling the use of GCM ciphers with BouncyCastleJsseProvider in FIPS mode. For use by those who are certain their cryptographic provider has FIPS validated support for GCM ciphers.

For developers who are using a provider with FIPS validated support for GCM ciphers, with this commit, you can set a Java system property `org.bouncycastle.jsse.fips.allowGCMCiphersIn12` to "true" which will add GCM ciphers as defined in [FipsUtils.java](https://github.com/bcgit/bc-java/blob/main/tls/src/main/java/org/bouncycastle/jsse/provider/FipsUtils.java)


dghgit:

PR #1868, opened 2024-10-15
Add getKeyIdentifier() methods to signature subpackets

vanitasvitae:


dghgit:
I've merged this one in, but I've also done some refactoring to move KeyIdentifier to the bcpg package - one thing to watch with bcpg is we should keep org.bouncycastle.openpgp dependencies out of it as people also use it stand-alone (or at least have in the passed, and knowing our luck still are...). A side effect of this is that KeyIdentifiers are now set up when a PGP public key is constructed as well, so getKeyIdentifier() should return the same object.

PR #1882, opened 2024-10-25
Fix PGPSecretKey.copyWithNewPassword() for AEAD

vanitasvitae:


vanitasvitae:
The fix is now part of #1926

PR #1926, opened 2024-11-28
High-Level API #0: Preparations for high-level OpenPGP Key Generator API

vanitasvitae:
This is part 1 of the split-up patch adding a high-level key generator API.
This one is focussed on changes required in `/openpgp/`, `/openpgp/operator/` and `/bcpg/`.

It also contains the fix from #1882 (the test will be included in the second PR).

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/KeyIdentifier.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/KeyIdentifier.java
@@ -1,6 +1,5 @@
 package org.bouncycastle.bcpg;
 
-import java.util.Iterator;
 import java.util.List;
 
 import org.bouncycastle.util.Arrays;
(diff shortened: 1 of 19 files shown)


vanitasvitae:
Some context for https://github.com/bcgit/bc-java/pull/1926/commits/32fc48b1af08b73120521fc9ee32e619da0e1b92 :
The Stateless OpenPGP Protocol allows implementations to emit the plain session key that was used to encrypt a message during the encryption process[1]. This feature can be used by implementations to retain access to the plaintext of ab encrypted message without the need to have access to decryption secret key material[2].

The best way I found to add support for session-key extraction in a backwards-compatible manner was by using a callback. Let me know what you think of this feature.

In https://github.com/bcgit/bc-java/pull/1926/commits/15ae18e3e61724d7b9b9a5401736ea6305d9dc7a I also exposed the encrypted data packet object. This is needed in order to emit, which encryption method (SEIPDv1, SEIPDv2, legacy SED, LibrePGP OED) was used to encrypt the message with.

[1]: https://www.ietf.org/archive/id/draft-dkg-openpgp-stateless-cli-11.html#section-5.4.3-16
[2]: https://www.ietf.org/archive/id/draft-ietf-lamps-e2e-mail-guidance-11.html#section-9.1-4.3

ligefeiBouncycastle:
There are some classes that are still under review and will be merged into the main branch later.

PR #1931, opened 2024-12-03
Fix symmetric v6 message encryption

vanitasvitae:
Commits 7d95b08316f0edf5773b0ea062694ed4b0e2b6ef and ea316319d029e9e1215c44e980509b3bf898cf05 introduced an error where the plain session key was passed in the wrong format causing the session-key wrapper to fail due to an invalid block size.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/PGPEncryptedDataGenerator.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/PGPEncryptedDataGenerator.java
@@ -224,10 +224,13 @@ public class PGPEncryptedDataGenerator
 
         boolean directS2K = !forceSessionKey && methods.size() == 1 &&
             methods.get(0) instanceof PBEKeyEncryptionMethodGenerator; // not public key
+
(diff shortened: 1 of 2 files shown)


ligefeiBouncycastle:
- As stated in [RFC 9580, Section 3.7.2.1, Table 2](https://www.rfc-editor.org/rfc/rfc9580.html#section-3.7.2.1), the Encryption column for 253 (AEAD) specifies the format as AEAD(HKDF(S2K(passphrase), info), secrets, packetprefix). The session key without checksum as part of its structure.
-  The sessionInfo is a constructed representation where a 1-byte algorithm identifier is prepended to the sessionKey, followed by a 2-byte checksum. In the original implementation of PBEKeyEncryptionMethodGenerator.generateV6ESK, there was a step to strip these additional 3 bytes from sessionInfo before usage.
To optimize the process and align with the documentation, the decision was made to use the sessionKey directly for version 6 packages (v6 package), bypassing the intermediate sessionInfo structure.

vanitasvitae:
Hm, apparently I forgot to implement PKESKv6 generation completely :D
I'll work on a fix.

PR #1938, opened 2024-12-09
Actually implement v6 PKESK encryption

vanitasvitae:
This PR
* fixes encoding of anonymous PKESKv6 packets
* actually implements encryption of messages using PKESKv6 packets


ligefeiBouncycastle:
@vanitasvitae, thank you for this PR! I have a couple of questions:

1. Why does the PublicKeyKeyEncryptionMethodGenerator.generateV3 function require parameters (int, byte[]) instead of just (byte[]), considering that encAlgorithm is not used?
2. Could we combine the following functions to simplify the code:
- Merge PublicKeyKeyEncryptionMethodGenerator.generateV3 and generateV6 into a single function.
- Similarly, unify PBEKeyEncryptionMethodGenerator.generateV4, generateV5, and generateV6.
This way, PGPKeyEncryptionMethodGenerator could have a single generate method.  (PGPKeyEncryptionMethodGenerator might later be converted to an interface).
Let me know your thoughts. Thank you.

ligefeiBouncycastle:
@vanitasvitae, thank you for this PR! I have one question:

1. Why does the PublicKeyKeyEncryptionMethodGenerator.generateV3 function require parameters (int, byte[]) instead of just (byte[]), considering that encAlgorithm is not used?

vanitasvitae:
> If encryptV3SessionInfoWithX25519X448Key and encryptV6SessionInfoWithX25519X448Key can be merged into one function, encryptSessionInfoV3 and encryptSessionInfoV6 could be merged into one function as well, with one additional change in generateV6 that get sessionInfoWithoutAlgId before calling encryptSessionInfoV6. How do you think about it?

I was able to merge some methods, reducing the LOCs a bit.

ligefeiBouncycastle:
@vanitasvitae Here is my understanding of how sessionInfo is used in relation to SymmetricKeyEncSessionPacket and PublicKeyEncSessionPacket. Please let me know if anything is incorrect. Thank you!
sessionInfo = algorithm identifier (1 byte) + session key + check sum (2 bytes).
In SymmetricKeyEncSessionPacket, The sessionIno needs:
- v4: Algorithm identifier + session key 
- v5: session key
- v6: session key
In PublicKeyEncSessionPacket, The sessionInfo needs:
- v3: 
  - Algorithm identifier + session key + checksum (except X25519 and X448)
  - Algorithm identifier + session key (for X25519 and X448)
- v6:
  - session key + checksum (except X25519 and X448)
  - session key (for X25519 and X448)

PR #1943, opened 2024-12-13
Make CompositePrivateKey::getAlgorithm() consistent with CompositePublicKey::getAlgorithm()

Honzaik:
Minor modification. A composite private key returns an OID as a result of getAlgorithm() but a composite public key returns a "human readable name."


dghgit:
Thanks for the patch. Merged!

PR #1966, opened 2025-01-13
Abstract core KEM functionality out of DHKEM 2/2

jmcrawford45:
This change was missed in the initial port of #1664. Without this change, customers cannot provide a non-DH KEM


dghgit:
It doesn't compile.

PR #2026, opened 2025-03-14
reduce number of Error Prone warnings and some other IDE warnings

winfriedgerlach:


dghgit:
Okay, ErrorProne is finally satisfied, at least until they add more. I think we have an approach for keeping on top of them now. Closing as done (temporarily, it'll be an endless battle...). Will look at the other noise-inducing issues as well.

PR #2027, opened 2025-03-14
fix bug in SecretKeyUtil

winfriedgerlach:
found a bug (ID string was added to map, not the ASN1ObjectIdentifier), but is this class even used?

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/util/SecretKeyUtil.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/util/SecretKeyUtil.java
@@ -11,28 +11,28 @@ import org.bouncycastle.util.Integers;
 
 public class SecretKeyUtil
 {
-    private static Map keySizes = new HashMap();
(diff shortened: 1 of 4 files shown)


dghgit:
Thanks for the patch, merged with a few revisions. Yes, it looks like the class has reached its used-by date. I've left it in as other people might be relying on it, but I've deprecated it. We'll see what shows up.

winfriedgerlach:
@dghgit Hi, thanks for you quick reaction! I don't see a change in SecretKeyUtil yet, ist there another branch that is not visible to regular users?
Also another question, I found *many* places where types could be added to generic classes and Java 8 features could be used. Is it of value to you if I provide PRs for that?

dghgit:
Somehow I missed the follow up. The answer is "possibly" I know that sounds a bit weird, but we're still doing a Java 5 to Java 8 release, so Java 8 features can still result in code duplication, which we try to avoid where we can.

winfriedgerlach:
All right, I understand that. Will only make suggestions that are Java 5 compatible.
At some point in time in the future, users will understand that you are dropping pre-Java-8, Java 8 is already 11 years old ;-)

PR #2032, opened 2025-03-19
fix: correct AsconCXof128 impl and add tests with ascon-c vectors

zhsnew:
The AsconCXof128 algorithm now correctly processes Z0 data, with test code added and validated using ascon-c test vectors.


ligefeiBouncycastle:
Thank you for This PR. The code has been merged.

PR #2041, opened 2025-03-27
SHA1Digest should not be cloned to MD5Digest

winfriedgerlach:
Maybe I'm wrong, but it looks like a bug to me if cloning a "SHA-1" digest results in an Md5Digest


dghgit:
Yes, I don't think that's right at all. I've made the change. Thanks for the report.

PR #2042, opened 2025-03-27
provide parametrized types for generics

winfriedgerlach:
The use of generics in the BC code base varies a lot. In many cases, raw types are still used. I think they can be parameterized without causing harm to existing code bases.

What do you think? If you are interested, I can provide a large PR with changes like the ones in this example PR.


peterdettman:
Modernization efforts commonly run afoul of the need to still compile for ancient java versions. In this _particular_ case I think the way we compile allows generic parameters to be present, but casting may still be required e.g. when returning an object from a collection, even with the parameters. I don't completely understand it, but I would say that it's probably not worth the effort (yet).

PR #2043, opened 2025-03-27
use for-each instead of enumerations

winfriedgerlach:
David put in the effort and made `ASN1Sequence` iterable. But many places in the code still use the old `Enumeration` access pattern. Would you accept changing these to `for-each` where possible? (better readability, elements are parameterized as `<ASN1Encodable>` which improves type safety)


dghgit:
Thanks for the PR. We're still shipping Java 4, slowly migrating to more modern styles as overlays kick in. Closing as partially merged and on-going.

PR #2106, opened 2025-06-16
Scout/java 5 features

winfriedgerlach:
Here is the first of some modernization PRs


silviuburceadev:
I'm just wondering if one needs to worry about thread safety on the switch from StringBuffer to StringBuilder in all these classes.

winfriedgerlach:
I was careful and only changed from `StringBuffer` to `StringBuilder` when the result of the StringBuilder was "consumed" in the same method - i.e. `sb.toString()` was called.

In cases where something else was done with `StringBuffer` (e.g., stored in a field, returned from method etc.) I didn't touch it.

PR #2107, opened 2025-06-16
Use enhanced for instead of while loops

winfriedgerlach:
Here is the 2nd of some modernization PRs


dghgit:
Thanks for the PR. We're still shipping Java 4, slowly migrating to more modern styles as overlays kick in. Closing as partially merged and on-going. This one would need to be rebased as well, had a crossed branch issue...

PR #2108, opened 2025-06-16
Use enhanced for loops instead of regular for loops part 1

winfriedgerlach:
Here is the 3rd of some modernization PRs


dghgit:
This one may take a while to deal with. I hate to admit this, but we actually still ship for Java 4, which believe it or not is still in use and in some environments the only option.

winfriedgerlach:
All right, I understand...

dghgit:
Thanks for the PR. We're still shipping Java 4, slowly migrating to more modern styles as overlays kick in. Closing as partially merged and on-going. This one would need to be re-based as well, had a crossed branch issue... I wouldn't worry about it though, in some universes time move much slower than it does here.

PR #2123, opened 2025-07-10
OpenPGP: Mitigate some RuntimeExceptions uncovered using Fuzzing

vanitasvitae:
Hey!

I did run junit-jazzer agains some PGPainless functions and uncovered some bugs in the bcpg codebase.

The commits in this PR harmonize the way malformed or strange packets are handled, by preventing runtime exceptions, such as IndexOutOfBounds, NegativeArraySize, IllegalArgumentExceptions and OutOfMemory errors.

I'm only just getting started with fuzzing, but I can highly recommend it!

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/UserAttributeSubpacketInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/UserAttributeSubpacketInputStream.java
@@ -13,6 +13,8 @@ public class UserAttributeSubpacketInputStream
     extends InputStream
     implements UserAttributeSubpacketTags
 {
+    public static int MAX_LEN = 5 * 1024 * 1024;
+
     InputStream in;
 
     public UserAttributeSubpacketInputStream(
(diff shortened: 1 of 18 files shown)


ligefeiBouncycastle:
Hi @vanitasvitae,
Thanks for your valuable PR in hardening the PGP packet handling against malformed inputs. Your fuzzing work is much appreciated.

Regarding the UserAttributeSubpacketInputStream.MAX_LEN setting (currently 5MB):
- We should revisit this limit – a prior fix in the C# library ([PR #638](https://github.com/bcgit/bc-csharp/pull/638)) encountered a similar issue where packet lengths could theoretically reach 4,082,119,084 bytes (~3.8GB). The resolution there set the max to 2³⁰ bytes (1,073,741,824 bytes)

ligefeiBouncycastle:
While 2³⁰ bytes (~1GB) still feels excessive for real-world use cases, it’s a practical upper bound and certainly more sane than the 2⁶⁴ max.

I'm fine with leaving it as a public static int, as I don’t expect most users to have a meaningful sense of what a "safe" maximum packet size should be. Exposing it allows advanced users to override the limit if necessary.

If there’s a configuration mechanism in place (e.g., via Properties), happy to align with that pattern if it improves usability and consistency.

I updated the logic in UserAttributeSubpacketInputStream.readPacket() to check StreamUtil.flag_eof before evaluating bodyLen, since a true flag_eof may validly result in a bodyLen of -1. This avoids potential false positives when handling EOF-flagged packets.

I've sent the branch for the final review. Let me know if there is any more update. Thank you.

peterdettman:
Please do not add mutable statics. If some limit is warranted here, use org.bouncycastle.util.Properties class for configuration.

However limit checking should begin at the top/outermost level with a constraint on the total size of the input. Low-level classes could be optimized to spot that their length is greater than the remaining limit. It looks like SignatureSubpacketInputStream started to head in that direction.

pg/src/main/java/org/bouncycastle/openpgp/PGPObjectFactory.java
dghgit, line 1: Is it possible to catch and rethrow inside PGPEncryptedDataList? It would be "less surprising" amongst other things if that could be done.

pg/src/main/java/org/bouncycastle/bcpg/UserAttributeSubpacketInputStream.java
dghgit, line 1: bobyLen checks should be after flag checks, have moved.

PR #2165, opened 2025-09-18
Adapted tls to support custom certificate types.

xipki:
This adaption allows one to support new certificate type, e.g. the C509 certificate (https://datatracker.ietf.org/doc/draft-ietf-cose-cbor-encoded-cert/), without modifying the bouncycastle classes.


xipki:
@peterdettman Hi Peter, could you please check this PR? We need this PR to demonstrate the integration of C509 certificates in the IETF COSE group.

xipki:
@gselander Here the PR.

PR #2169, opened 2025-09-27
OpenPGPKeyReader: Use consistent method name parseKeyOrCertificate()

vanitasvitae:
Added new methods as replacement for mis-named 'parseCertificateOrKey()' methods and deprecated old methods


dghgit:
Thanks for the patch. Merged.

PR #2178, opened 2025-10-13
Fix NPE when testing passphrase for stripped secret keys

vanitasvitae:
Fixes https://github.com/bcgit/bc-java/issues/2173


dghgit:
Thanks for the patch. Now merged and on https://downloads.bouncycastle.org/betas

PR #2184, opened 2025-10-21
OpenPGP v3 key: Fix NPE in PGPSignature.encode() for missing subpacket areas

vanitasvitae:
V3 keys might have signatures with missing subpacket areas. This is especially true for OpenPGP v3 keys.
This patch fixes an NPE in PGPSignature.encode() for missing subpacket areas.


vanitasvitae:
Merged into #2171

PR #2185, opened 2025-10-21
Fix ClassCastException when trying to verify misplaced subkey binding…

vanitasvitae:
… signatures

Misplaced subkey binding signatures could result in the Signer being initialized using the wrong key material,
causing ClassCastExceptions.
This patch catches those exceptions and rethrows them as gracefully handled PGPSIgnatureExceptions.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/openpgp/api/OpenPGPCertificate.java
+++ b/pg/src/main/java/org/bouncycastle/openpgp/api/OpenPGPCertificate.java
@@ -1949,7 +1949,7 @@ public class OpenPGPCertificate
                     throw new IncorrectOpenPGPSignatureException(this, "Key Signature is not correct.");
                 }
             }
-            catch (PGPException e)
+            catch (PGPException | ClassCastException e)
             {
                 this.isCorrect = false;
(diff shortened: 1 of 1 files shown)


vanitasvitae:
To add more details:
The bug is hit when BC tries to verify validity of an OpenPGP certificate with the following format:
```
PrimaryKey[Ed25519] 0x1234
  Subkey[Ed25519] 0xABCD
  Subkey[ECDH] 0x9876
    SubkeyBindingSignature[Ed25519, by 0x1234] (misplaced, actually belongs to 0xABCD)
      Embedded PrimaryKeyBindingSignature[Ed25519, by 0xABCD]
```
When verifying the validity of 0x9876, the misplaced binding signature is evaluated. BC tries to initialize the signature, and since the key material of 0x9876 is neither an instance of `Ed25519PrivateKeyParameters` nor of `Ed25519PublicKeyParameters`, it is assumed to be Ed448 key material: https://github.com/bcgit/bc-java/blob/main/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcImplProvider.java#L101-L105
The Ed448 signer then throws a `ClassCastException` when trying to cast the ECDH key material.

PR #2205, opened 2025-11-27
Blake2bDigest: defensive improvements and cleanup

lomigmegard:
While looking at `Blake2bDigest` implementation I noticed a few things that could be tidied up.

Defensive changes:
- Validate key length before copying to fail fast and avoid unnecessary allocation on invalid input
- Centralize state cleanup in reset() to avoid double-zeroing and ensure consistent cleanup path

Code quality:
- Use explicit 64-bit literals for long field `f1` for clarity. I initially thought it was a bug, but Java bit extension makes the current syntax correct but misleading.
- Other trivial cleanup

LLM disclaimer:
- A human, myself, wrote all changes by hand
- A LLM reviewed the changes and created the commit message

### DIFF
--- a/core/src/main/java/org/bouncycastle/crypto/digests/Blake2bDigest.java
+++ b/core/src/main/java/org/bouncycastle/crypto/digests/Blake2bDigest.java
@@ -33,7 +33,7 @@ import org.bouncycastle.util.Pack;
 
 
 /**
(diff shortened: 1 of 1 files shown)


peterdettman:
Merged, thanks for the PR.

PR #2233, opened 2026-01-19
OpenPGPKeyGenerator: Add factory methods for OpenPGP keys based on RSA

vanitasvitae:
This PR adds two factory methods for generating RSA-based OpenPGP keys.


dghgit:
Thanks for the patch, now merged.

PR #2236, opened 2026-01-21
OpenPGPCertificate: Allow getting encryption keys by purpose

vanitasvitae:
The OpenPGP protocol defines two different encryption key flags: ENCRYPT_STORAGE and ENCRYPT_COMMS.

This PR adds methods to select encryption keys of a certificate by passing in key flags as array of masks.
This allows the user to encrypt a message for a specific use-case, such as data in transmission or data at rest.


vanitasvitae:
This got merged as fcedfe315c4c83626b51f0c9f7bd8f2d80b3e88d

Issue #1481, opened 2023-08-25
NameConstraints intersectPermittedSubtree is not working when name constraints extensions are set in multiple place in the CA hierarchy.


kushshrestha01:

kushshrestha01:
For example: if you have ```.permitted1.permitted2.com``` in the RootCA and SubordinateCA then the ```intersectDNS``` will return null instead of ```.permitted1.permitted2.com```

```
    public void intersectPermittedSubtree(GeneralSubtree[] permitted)
    {
        Map subtreesMap = new HashMap();

        // group in sets in a map ordered by tag no.
        for (int i = 0; i != permitted.length; i++)
        {
            GeneralSubtree subtree = permitted[i];
            Integer tagNo = Integers.valueOf(subtree.getBase().getTagNo());
            if (subtreesMap.get(tagNo) == null)
            {
                subtreesMap.put(tagNo, new HashSet());
            }
            ((Set)subtreesMap.get(tagNo)).add(subtree);
        }

        for (Iterator it = subtreesMap.entrySet().iterator(); it.hasNext();)
        {
            Map.Entry entry = (Map.Entry)it.next();

            // go through all subtree groups
            int nameType = ((Integer)entry.getKey()).intValue();
            switch (nameType)
            {
            case GeneralName.otherName:
                permittedSubtreesOtherName = intersectOtherName(permittedSubtreesOtherName,
                    (Set)entry.getValue());
                break;
            case GeneralName.rfc822Name:
                permittedSubtreesEmail = intersectEmail(permittedSubtreesEmail,
                    (Set)entry.getValue());
                break;
            case GeneralName.dNSName:
                permittedSubtreesDNS = intersectDNS(permittedSubtreesDNS,
                    (Set)entry.getValue());
                break;
            case GeneralName.directoryName:
                permittedSubtreesDN = intersectDN(permittedSubtreesDN,
                    (Set)entry.getValue());
                break;
            case GeneralName.uniformResourceIdentifier:
                permittedSubtreesURI = intersectURI(permittedSubtreesURI,
                    (Set)entry.getValue());
                break;
            case GeneralName.iPAddress:
                permittedSubtreesIP = intersectIP(permittedSubtreesIP,
                    (Set)entry.getValue());
                break;
            default:
                throw new IllegalStateException("Unknown tag encountered: " + nameType);
            }
        }
    }
    private Set intersectDNS(Set permitted, Set dnss)
    {
        Set intersect = new HashSet();
[...]

PR #2244, opened 2026-01-27
Fix Strings.split() to handle delimiters at position 0

diresh-shrestha:
Fixes: https://github.com/bcgit/bc-java/issues/1481

The split method incorrectly treated delimiters at index 0 as 'no delimiter found' because the condition used 'tokenLocation > 0' instead of '>= 0'.

This caused issues in PKIXNameConstraintValidator.withinDomain() when processing DNS name constraints with leading dots (e.g., '.example.com'). The method would return the entire string as a single element instead of properly splitting it, causing certificate validation to fail when both Root CA and Subordinate CA had the same dNSName in NameConstraints.

Fix: Change 'tokenLocation > 0' to 'tokenLocation >= 0'

Before: Strings.split(".permitted", '.') -> [".permitted"]
After:  Strings.split(".permitted", '.') -> ["", "permitted"]

### DIFF
--- a/core/src/main/java/org/bouncycastle/util/Strings.java
+++ b/core/src/main/java/org/bouncycastle/util/Strings.java
@@ -333,7 +333,7 @@ public final class Strings
         while (moreTokens)
         {
             int tokenLocation = input.indexOf(delimiter);
-            if (tokenLocation > 0)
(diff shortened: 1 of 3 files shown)


dghgit:
Thanks for the patch, now merged.

Issue #2246, opened 2026-01-29
Main branch not buildable after checkout


hendrikebbers:
Current main branch can not be build on my machine. It looks like a lot of tests are failing based on java.io.FileNotFoundException. Sadly the output does not give me a hint about the problem.

I'm running Java 21:

```
$ java -version
openjdk version "25.0.1" 2025-10-21 LTS
OpenJDK Runtime Environment Temurin-25.0.1+8 (build 25.0.1+8-LTS)
OpenJDK 64-Bit Server VM Temurin-25.0.1+8 (build 25.0.1+8-LTS, mixed mode, sharing)
```

Gradle log:

```
$  ./gradlew clean build

> Configure project :
Environment setup:
Looking for JDK ENV 'BC_JDK8' found  null
Looking for JDK ENV 'BC_JDK11' found  null
Looking for JDK ENV 'BC_JDK17' found  null
Looking for JDK ENV 'BC_JDK21' found  null
Looking for JDK ENV 'BC_JDK25' found  null

> Task :core:compileJava
...... Removed all the compile output



> Task :core:test

AllTests > org.bouncycastle.pqc.math.ntru.test.HPSPolynomialTest.testR2Inv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HPSPolynomialTest.testS3Inv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HPSPolynomialTest.testRqInv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HPSPolynomialTest.testLift FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HRSSPolynomialTest.testR2Inv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HRSSPolynomialTest.testS3Inv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HRSSPolynomialTest.testRqInv FAILED
    java.io.FileNotFoundException at null:-1

AllTests > org.bouncycastle.pqc.math.ntru.test.HRSSPolynomialTest.testLift FAILED
    java.io.FileNotFoundException at null:-1

[...]

PR #2248, opened 2026-01-30
Better handling for test data

hendrikebbers:
Suggestion to make the solution for https://github.com/bcgit/bc-java/issues/2246 more clear and add a way to place test data on a different folder (by using an ENV).

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ export BC_JDK17=/path/to/java17
 
 ## Building
 
+To run the tests of the project as part of the build test data is needed.
+Our test data can be found at the https://github.com/bcgit/bc-test-data.git repository.
+Best approach to provide the data for the tests is to check out the repository and define the `BC_TEST_DATA_DIR` env variable pointing to that folder.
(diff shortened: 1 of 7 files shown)


hendrikebbers:
Did the mistake first, realized it later but did not updated the comment ;D

dghgit:
I've updated the README, it's a good idea. My instinct is to go no further than that, I'll let this sit for a bit while I think on it further.

dghgit:
Okay, so starting to work through the Round 3 signature algorithms and bc-test-data is starting to look like Eric Cartman after a month on Weight Gain 4000... I don't think the simple directory search is going to be enough in some environments, so I think this is an idea whose time has definitely arrived. I've added the changes with some modifications - historically we've always used "bc.test.data.home" as the location, and some ant scripts still use (or at least did).

I've mentioned the "new way" in a couple of places in README.md, but "Building overview" is probably a good place to look.

Closing now as done. Re-open and let me know if things still look unclear.

Issue #2239, opened 2026-01-23
Always add exception cause


hendrikebbers:
Today Bouncy Castle uses several different ways/pattern how exceptions are handled. In general the best practice would be code like this:

```
try {
    doSomeMagic(); 
} catch(AnyException e) {
    throw new ConcreteException("Error in doing some magic!", e);
}
```

While many places of Bounce Castle follow that pattern some places are use other patterns like this:

```
try {
    doSomeMagic(); 
} catch(AnyException e) {
    throw new ConcreteException("Error in doing some magic:" + e.getMessage());
}
```

By doing so the original exception (including any possible cause and stack trace) got lost. Especially for application in production that relay on logging regarding errors, those information are very important to understand the cause of an exception.

There are even some other patterns in the code like this snippet from [DerUtil.java](https://github.com/bcgit/bc-java/blob/09212a09769b6cca130c9d54f9c27d264e6d7531/core/src/main/java/org/bouncycastle/crypto/util/DerUtil.java#L30):

```
static byte[] toByteArray(ASN1Primitive primitive)
    {
        try
        {
            return primitive.getEncoded();
        }
        catch (final IOException e)
        {
            throw new IllegalStateException("Cannot get encoding: " + e.getMessage())
            {
                public Throwable getCause()
                {
                    return e;
                }
            };
        }
    }
```

A good starting point to find the places that does not use the best practice pattern is by [searching in the project for `e.getMessage()`](https://github.com/search?q=repo%3Abcgit%2Fbc-java%20e.getMessage()&type=code).

I would be more than happy to work on this issue and migrate all exception handling in Bounce Castle to the same pattern. If you agree on this please assign me to the issue and leave a comment :)

dghgit:
IllegalStateException and IllegalArgumentException have now been addressed.

dghgit:
IOException and GeneralSecurityException have now been dealt with as well.

PR #2250, opened 2026-02-01
Always add exception cause (first PR)

hendrikebbers:
This PR is based on https://github.com/bcgit/bc-java/issues/2239

The PR does not fix the issue but only changes some first places. Once the PR is merged I will continue by creating PRs for each module.


dghgit:
This has been applied using the Exceptions class. New behavior is in the version on https://www.bouncycastle.org/betas

Issue #2234, opened 2026-01-20
DeltaCertificateRequestAttributeValueBuilder cannot encode only deltas differences as per draft-bonnell-lamps-chameleon-certs §5.1


coeditor600:
Hello Bouncy Castle team,
I’m currently working with the org.bouncycastle.asn1.x509.DeltaCertificateDescriptor and related classes introduced for Chameleon Certificates. While reviewing and using the Implementation, I noticed following problem:

### Problem
The class DeltaCertificateRequestAttributeValueBuilder does not provide a mechanism to encode only the differences between the base certificate and the delta certificate, as described in Section 5.1 of the Chameleon Certificates draft:

Internet-Draft:
https://datatracker.ietf.org/doc/draft-bonnell-lamps-chameleon-certs/
Relevant section: §5.1 – Creating a Certificate Signing Request for Paired Certificates

According to the draft, the Delta certificate request is intended to encode only the difference to the base certificate. However, the current builder API requires explicitly setting all fields and does not support automatic diffing against a base certificate.

### Impact

* Increases risk of non-compliant outputs
* Forces external logic to implement the diffing between base and delta certificates

### Proposed Improvement
I have already implemented a draft solution that:

* compares a base certificate and delta certificate
* encodes only fields whose values differ
* produces a DeltaCertificateDescriptor aligned with §5.1 of the draft

I would like to contribute this enhancement to this project, via a new builder method or constructor that takes a base certificate or a dedicated “delta-only” encoding API.

### Questions

Is this limitation already known?
Would the maintainers be open to a pull request implementing delta-only encoding?
Is there a preferred design direction (API shape, naming, packages)?

I’m happy to open a PR once I understand the preferred approach.

Thank you for your work on Bouncy Castle and for reviewing this issue.

dghgit:
Now fixed and in https://www.bouncycastle.org/betas

PR #2259, opened 2026-02-18
DeltaCertificateRequestAttributeValueBuilder cannot encode only differences compared to the base certificates

coeditor600:
This is a PR to fix the chameleon certificate differences mentioned in Issue #2234


ligefeiBouncycastle:
Thanks for the PR! Now the code has been merged into the main branch, and later it will be included in [https://www.bouncycastle.org/betas](https://www.bouncycastle.org/betas).

PR #2272, opened 2026-03-12
EC: Validate local EC domain parameters for MQV key agreement

Joowon-Seo:
Add validation during MQV key agreement initialisation to ensure
the local static and ephemeral EC keys share the same domain parameters.

Previously these parameters were not explicitly validated, allowing
mismatched keys to proceed until later stages of the agreement.

Tests added for matching and mismatched curve scenarios.


peterdettman:
The proposed test case `ECKeyAgreementDomainValidationTest` does not test ECMQV, and already passes without the proposed changes in `KeyAgreementSpi`. Those `KeyAgreementSpi` changes are also redundant with checks done by `new MQVPrivateParameters(staticPrivKey, ephemPrivKey, ephemPubKey)` (The TODO seems a bit obsolete).

It may be worth adding a test case to `org.bouncycastle.jce.provider.test.MQVTest` confirming the existing checks already work, and also reviewing whether all the other-party keys are also checked when they are available.

PR #2277, opened 2026-03-27
Hash2curve

Razumain:
Implementation of RFC 9380 Hashing to Elliptic Curves as well as Hash to scalar according to RFC 9497.


peterdettman:
This has now been merged and will be included in the imminent 1.84 release. Thanks for the PR!

PR #2285, opened 2026-04-21
Added license prefix to help self-identify it

j256:


dghgit:
Thanks for the suggestion, this has been added.

PR #2290, opened 2026-04-30
Make KEM abstract methods protected to allow external subclassing

jmcrawford45:
Package-private abstract methods on KEM prevent users from implementing custom KEMs (e.g. XWing) outside the org.bouncycastle.crypto.hpke package, requiring them to place subclasses inside the BC package as a workaround.

Promote all abstract methods on KEM from package-private to protected. Update DHKEM.getEncryptionSize() to match — it was the only remaining package-private override; all other DHKEM overrides were already public or protected.

This is a source-compatible change: existing subclasses (DHKEM) that override with broader visibility (public/protected) are unaffected.


dghgit:
I'm not sure if this is really a good idea, I think you may, understandably, be looking at the wrong class for the job - we've found a better way (I'll admit the term is always a bit relative...) of describing KEM's which also works with the JCE, including the new KEM API which has been backported to Java 17. Would you have a look at the XWing implementation in core/src/main/java/org/bouncycastle/pqc/crypto/xwing and let me know what you think?

PR #2298, opened 2026-05-09
Update README for JDK25.

msgilligan:
The build now seems to work on (and require) JDK25 and the BC_JDK21 environment variable should be
added to the list of env variables.


dghgit:
Thanks for bringing it up. Update now merged.

PR #2309, opened 2026-05-27
Fix the issue where specific exception details were ignored.

wolf-hunter404:


dghgit:
Thanks for the suggestion. I've added causation setting for these exceptions and a few others, "merged with revision".

PR #2315, opened 2026-06-05
disable DTDs and external entities in KMIPInputStream

digi-scrypt:
1. KMIPInputStream builds its XMLEventReader from a bare XMLInputFactory, so a DOCTYPE in a parsed KMIP message is honoured and external SYSTEM entities get resolved as parse() pulls events.
2. a file:// or http:// entity then turns into a local file read or an outbound request from whichever side parses the message; aiming one at a missing path drops a "No such file or directory" straight out of the parse loop, which is the tell.
Set SUPPORT_DTD=false (and external entities off as a backstop) on the factory before creating the reader. Messages without a doctype parse exactly as before.


dghgit:
Thanks for the patch. Merged!

PR #2316, opened 2026-06-09
verify mls membership and confirmation tags in constant time

rootvector2:
found by auditing mls MAC verification: `PublicMessage.unprotect` and `Group` compared a secret-keyed HMAC membership/confirmation tag against the wire value with early-exit `Arrays.areEqual`/`Arrays.equals`, a byte-by-byte timing oracle reachable from `Group.handle(byte[])`, so all three sites now use `Arrays.constantTimeAreEqual` like the rest of the library.


dghgit:
Thanks for the patch, merged with minor revisions. Now on https://www.bouncycastle.org/betas

PR #2317, opened 2026-06-09
PGPPBEEncryptedData: Expose keyData member

vanitasvitae:
PGPPBEEncryptedData: Expose keyData member in order to allow downstream parameter validation

Downstream applications like PGPainless may want to validate SKESK parameters, such as S2K parameters against application policies.
One notable example would be the Argon2 memoryExponent parameter, which - if not guarded - could result in OOM errors for large exponents.


vanitasvitae:
Closed, as Argon2 parameters are gated through https://github.com/bcgit/bc-java/commit/9a8ee882ca9ab0c9d93158ee3990c4b3ade80c77

PR #2322, opened 2026-06-11
Move Argon2 memory size exponent check to PGPPBEEncryptedData.checkPa…

vanitasvitae:
…rameterBounds()

This moves the parameter validation out of the packet parsing step into the decryption step.
The exception changed from an IOException to PGPException

Relates to #2283


dghgit:
Thanks for the patch. Merged with minor revisions. Now on https://www.bouncycastle.org/betas

PR #2325, opened 2026-06-13
fix defeated invalid-char check in ArmoredInputStream.decode

rootvector2:
`ArmoredInputStream.decode` masks the two decoding-table lookups with `& 0xff` in the two-pad (`XX==`) branch, so the `0xff` sentinel a non-alphabet character maps to becomes 255 and the `(b1 | b2) < 0` guard right below it can never fire. Spotted by diffing the three branches: a final group like `!!==` is decoded to a junk byte while `!!!=` and `!!!!` correctly throw `invalid armor`, and `Base64Encoder` rejects the same input. Dropping the mask makes the 2-char ending validate like the one-pad and no-pad cases.


dghgit:
Thanks for the patch. Merged!

PR #2326, opened 2026-06-15
use Files.createTempFile for s/mime backing temp files

rootvector2:
`SMIMEUtil` and `SMIMESignedParser` spool decrypted, decompressed and signed S/MIME content to a backing file created with `File.createTempFile`, which honors the umask and lands at mode `0644` on a normal POSIX box. The spooled bytes are plaintext, so on a shared host any local user can read a decrypted message out of the temp dir before it gets cleaned up. Found while auditing the temp-file paths in the mail module.

Switch the four sites to `Files.createTempFile`, which creates the file `0600` on POSIX and owner-only on Windows. Keeping the restriction in the helpers means every caller going through `toMimeBodyPart` / `toWriteOnceBodyPart` / the parser temp file gets it without each one re-deriving the check.


dghgit:
Thanks for the patch. Merged!

PR #2329, opened 2026-06-18
ArmoredInputStream: CSF - Fix handling of dash-prefixed messages

vanitasvitae:
This PR fixes the way BC handles messages created using the signature framework (csf) that contain malformed prefixed dashes.

When creating a cleartext signed message, all lines in the payload that start with a dash (`-`) need to be prefixed with `- ` (dash minus).

The `ArmoredInputStream` does assume, that a line in a CSF signed message that starts with `-` is either a header if the second character is also an `-`, or it is a dash-escape with the second character being *blindly* assumed to be a space.

This leads to situations where a cleartext signature over "payload" also checks out if the CSF message payload is modified to "-Xpayload".

The PR fixes this behavior by checking whether the second character is a space. If not, it emits the line as is, or throws an exception if configured to do so.

It adds a builder method `.setRejectPrefixedDashesInCSFMessages()` that can be used to configure the `ArmoredInputStream` to reject such messages as malformed.

### DIFF
--- a/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
+++ b/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java
@@ -146,6 +146,7 @@ public class ArmoredInputStream
     boolean restart = false;
     StringList headerList = Strings.newList();
     int lastC = 0;
+    int lookAhead = -1;
     boolean isEndOfStream;
 
     private boolean validateAllowedHeaders = false;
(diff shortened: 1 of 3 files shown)


vanitasvitae:
I changed the behavior, such that malformed payload lines are now returned as they are if the rejection flag is set to false. Note: This is still different from the original (also faulty) behavior.
```
CSF message payload:
-Xpayload

ArmoredInputStream.read() before:
payload

ArmoredInputStream.read() now:
-Xpayload
```

Edit2:
Apparently, there is an edge case where GnuPG allowed creating CSF messages without dash escaping, see `--not-dash-escaped` [here](https://www.gnupg.org/documentation/manuals/gnupg-devel/GPG-Esoteric-Options.html).
This option has since been deprecated.

So this PR indeed fixes BCs behavior by rejecting such edge case messages by default, while allowing configuration to emit the intact, unescaped payload if desired.

dghgit:
Thanks for the patch. Merged!

PR #2331, opened 2026-06-19
reject empty sequence in size (1..max) x509 extension parsers

rootvector2:
the `SEQUENCE SIZE (1..MAX)` x509 extension parsers `CertificatePolicies`, `PolicyMappings`, `ExtendedKeyUsage`, `CRLDistPoint` and `SubjectDirectoryAttributes` accept an empty `SEQUENCE` from an untrusted certificate, unlike `AuthorityInformationAccess` which already rejects it, so each now applies the same `sequence may not be empty` check RFC 5280 requires.


dghgit:
Thanks for the patch. Merged!

PR #2339, opened 2026-06-28
Implement support for OpenPGP External Secret Keys

vanitasvitae:
See https://datatracker.ietf.org/doc/draft-dkg-openpgp-external-secrets/

This protocol extension aims to define ways to signal that an OpenPGP key has its private key material moved to an external place, e.g. a smart card or hardware token.

Compared to GnuPGs GNU_DUMMY_S2K, the external key mechanism will hopefully eventually be adopted as an official standard by the IETF.


dghgit:
The draft appears to have been allowed to expire. Do you know if this is happening else where?

vanitasvitae:
I do believe this is common for internet drafts. There has not been any activity on that draft for a while, so it expired after 6 months of inactivity. I do think, the document can become non-expired if the author publishes changes though.

I'll ask in the OpenPGP working group, what the next steps are.

vanitasvitae:
This has been merged as https://github.com/bcgit/bc-java/commit/f9a2bf1c449b6f96dbd046d1e170df43389c6f55

PR #2340, opened 2026-06-30
Fix bad numbering in BIP340Signer comment

liamgilligan:
Updates step numbering to match the actual number of steps being performed, as outlined in [bip-0340.mediawiki](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#user-content-Default_Signing). Note that these steps:
```
- Let rand = hashBIP0340/nonce(t || bytes(P) || m).
- Let k' = int(rand) mod n.
```
are folded into Step 6.


dghgit:
Thanks for noticing this. Merged with one small revision. I think the 9-10, 11-12 are okay, the off by one that occurred at step 1 got accounted for... Let me know if you think I'm confused.

dghgit:
Assumed patch okay.

PR #2342, opened 2026-07-04
avoid classcastexception in cms signerinfo decode via getinstance

rootvector2:
`SignerInfo`'s decode constructor cast the `version` and trailing `unsignedAttrs` elements instead of using `getInstance`, so a malformed-but-parseable CMS message leaks a `ClassCastException` past `CMSSignedData.verifySignatures()`'s `throws CMSException` contract when `getSignerInfos()` / `getCounterSignatures()` enumerate the signers outside the `getSignedData()` cast guard; found while checking siblings of the recent `asn1.cms.Attribute` type-check, fixed by decoding via `ASN1Integer.getInstance` / `ASN1TaggedObject.getInstance` to match the already-hardened `asn1.pkcs.SignerInfo`.


dghgit:
Thanks for the patch. Merged!

PR #2345, opened 2026-07-07
test: exercise SPHINCS+ OID round-trip for -simple variants

major:
This removes a stale `FIXME` in `SphincsPlusTest` that was skipping the SubjectPublicKeyInfo/PrivateKeyInfo (OID) encode/decode round-trip for SPHINCS+ "-simple" hash-mode variants.

**Why the FIXME was stale**: `sphincsPlusOids`/`sphincsPlusParams` in `pqc/crypto/util/Utils.java` already have correct forward and reverse OID mappings for all "-simple" variants (haraka has dedicated `_simple`-suffixed constants; sha2/shake reuse the unsuffixed constant, which now maps to the final SLH-DSA standard OID rather than the round-3-draft one). This was wired up in `729a3aa83` ("interop updates for SPHINCS+", Nov 2023), which aligned BC's naming with FIPS 205 finalizing on the "simple" hashing mode and dropping "robust". The test guard just never got removed after that change landed.

**What changed**:
- Removed the `// FIXME No OIDs for simple variants of SPHINCS+` comment and the `if (name.indexOf("-simple") < 0)` guard, so the OID round-trip now runs unconditionally for every variant.
- Added `assertTrue(... parameters == pubParams.getParameters())` / same for private key, so the test verifies `decode(encode(params))` returns the exact same parameter-set instance, not just that raw key bytes happen to match afterward.
- Applied identically to the `jdk1.3` mirror of this test.

**Verification**: ran `SphincsPlusTest` locally (via `junit.textui.TestRunner`, since the gradle module restricts `--tests` to `AllTest*`), all 9 cases pass, including every `-simple`, `-robust`, and haraka variant now exercising the full ASN.1 round-trip plus the new parameter-identity checks.

Investigation and fix drafted with AI assistance (Claude); I reviewed the git history, verified the OID mapping tables, and ran the tests myself before submitting.


dghgit:
Thanks for the patch. I've merged this, but please note SPHINCS+ will be getting deleted in the next release or two in favor of SLH-DSA. The new PQC algorithms take a considerable amount of space and with the new prototype signature algorithms around as well, it's becoming quite a challenge to stop the provider size exploding.

dghgit:
Separately, it looks like you might also need to rebase your branch. We had a cross commit branching problem...

PR #2346, opened 2026-07-07
fix off-by-4 header length guard in NotationData.verifyData

rootvector2:
`NotationData.verifyData` counts only `nameLength + valueLength + 4` against the body length while its own header is 8 octets (`flags[4] || nameLength[2] || valueLength[2]`), so a crafted critical notation subpacket that declares more name/value data than the body carries slips past the guard and then throws `ArrayIndexOutOfBoundsException` from `getNotationName()`/`getNotationValueBytes()` on the signature-verify path instead of a clean rejection; counting the full header rejects the short body at parse and completes the sibling `Features`/`TrustSignature`/`SignatureTarget`/`RevocationKey`/`RevocationReason` truncation guards; spotted while fuzzing the signature-subpacket parser.


dghgit:
Thanks for the patch. Merged!

PR #2347, opened 2026-07-09
OPS packet: Default to new packet format

vanitasvitae:
Hey!
Someone [noticed](https://codeberg.org/PGPainless/pgpainless/issues/33#issue-6191492) that BC still defaults to the old legacy packet format when emitting OnePassSignaturePackets.

This patch changes the default value to the new packet format.


dghgit:
I've uploaded a new jar to https://www.bouncycastle.org/betas - with the patch, I changed it as the original one is on the parser, I suspect just an accident. One thing, it defaults to new for V6, but legacy will still work okay, if you set PacketFormat.CURRENT it should run with the new style for everything though.

dghgit:
Closing, assumed fixed.

PR #2348, opened 2026-07-10
reject cr/lf in s/mime writer header names and values

rootvector2:
`SMIMEEnvelopedWriter.Builder.withHeader` and `SMIMESignedWriter.Builder.withHeader` store a header name/value as given, and `Headers.dumpHeaders` writes each one out followed by a single `CRLF`, with nothing rejecting a name or value that already contains `CR`/`LF`. an app that sets a header like `Subject` from user input can inject extra header lines, and a trailing `CRLFCRLF` forges a body ahead of the real part, i.e. MIME header injection. found while auditing the mime writers for the read-side hazard `SMIMEUtil.unfold` already strips.

validate the name and value in both `withHeader` methods through a new `SMimeUtils.checkHeader`, rejecting `CR`/`LF` with `IllegalArgumentException`. the guard belongs at that entry point because it is the only untrusted header source; the writers set their own headers directly, including the folded multipart `Content-Type` boundary that legitimately carries `CRLF`+tab, so valid output is unchanged.


dghgit:
Thanks for the patch. I'd say the risk is somewhat over blown here, I'd say it's a worthwhile hardening though, as it blocks a misuse of the API. I will say if this is the best that whatever tool you're using can come up with, I'm feeling quite relieved... at least for now... Now up on https://www.bouncycastle.org/betas

rootvector2:
agreed it's more hardening than exploit, blocking the misuse at the entry point was the intent. thanks for taking it into the betas.

PR #2349, opened 2026-07-11
Reset KGCM associatedText on reInit of initialAssociatedText

tonywasher:
If initialAssociatedText is specified for KGCM mode, it is immediately applied to associatedText, and the reset() method will automatically re-initialise the associatedText buffer correctly. 

However the associatedText is not cleared on a subsequent init(), leading to an incorrect associatedText buffer for that iteration. 
The buffer needs to be cleared on any init() call.


dghgit:
Thanks for the patch. Merged!

PR #2359, opened 2026-07-15
guard minimum length in sm2 decrypt and gost/dstu/desede/rc2 unwrap

rootvector2:
`SM2Engine.decrypt` and the `GOST28147`/`DSTU7624`/`DESede`/`RC2` key-wrap `unwrap` engines sized their output as `new byte[inLen - overhead]` without first checking `inLen` covers that overhead, so a short attacker-supplied ciphertext threw `NegativeArraySizeException`/`ArrayIndexOutOfBoundsException` instead of the declared `InvalidCipherTextException`; found auditing these against the already-guarded `IESEngine`/`RFC3394WrapEngine`/`RFC5649WrapEngine` and adds the same minimum-length check.


ligefeiBouncycastle:
Thank you for the PR, nice catch. A few days ago I had identified the same issue in SM2 and GOST28147, and your PR also covers DSTU7624, DESede and RC2, so this is a good opportunity to combine the fixes. I'll merge them together.

rootvector2:
sounds good, makes sense to combine them. thanks for picking it up.

rootvector2:
sounds good. `ShortWrapCipherTextTest` in this PR exercises all five engines, so it should carry over to the combined fix as-is.

ligefeiBouncycastle:
Thank you for this PR. It's now up on https://www.bouncycastle.org/betas if you want to try the revision.

PR #2373, opened 2026-07-28
test the returned type in oer getInstance identity guards

rootvector2:
the identity fast path in `UINT32.getInstance`, `etsi102941.basetypes.Version.getInstance` and `EtsiTs103097DataEncryptedUnicast.getInstance` tests a sibling type instead of the one it casts to (and `OEROptional.getObject` has the `isInstance` arguments transposed), so the first two reject their own type with `IllegalArgumentException` from `ASN1Integer.getInstance`, the third throws `ClassCastException` on an `EtsiTs103097DataEncrypted`, and the fourth never takes the cast path at all - found by turning on the `IsInstanceIncompatibleType` and `InstanceOfAndCastMatchWrongType` error-prone checks that `build.gradle` currently switches off.


dghgit:
Thanks for the patch. This has been merged.

One thing, you'll need to rebase your fork - we've had a dual lineage issue which has only recently surfaced. GitHub main is now the canonical lineage again.

rootvector2:
thanks for merging. noted on the lineage issue, I'll rebase the fork onto the canonical main.

PR #2374, opened 2026-07-29
PGSC: Add JCE implementation for OpenPGP Smart Card message decryption

vanitasvitae:
This PR is based on and includes #2339 

Inhere, I modified BCs API to allow for message decryption with external secret keys.
On top of that, I introduced `AbstractPublicKeyDataDecryptorFactory.getCryptoCallback()` and reworked the `BcPublicKeyDataDecryptorFactory` to cleanly separate out PK crypto operations during session-key recovery, for sake of simple delegation to hardware devices in subclasses.

Furthermode, the new module `pgsc` now contains the initial design of a smart card API for bcpg. It currently supports listing cards, uploading key material and decrypting messages using both a simulated smart card backend, as well as an implementation using the YubiKit library.

TODO:
* [ ] Apply changes to `JcaPublicKeyDataDecryptorFactory`
* [ ] Implement `JcaExternalPublicKeyDataDecryptorFactory`
* [ ] Consider reworking `YubikeyPublicKeyDataDecryptorFactory` to implement both a BC and JCA version.


ligefeiBouncycastle:
Hi @vanitasvitae,
Thanks for the JCE work,  it closes all three TODOs from the PR description, and I see you landed on the same conclusion we did about moving `PublicKeyCryptoCallback` out of AbstractPublicKeyDataDecryptorFactory.

Before going further, would you mind resyncing the branch onto current main and continuing from there? The earlier state of this PR is already merged, with a review pass on top, so working from current main means you can see what we've changed, and your own commits won't be masked by the ~4000 commits of drift. It also makes anything you need from our side easy to pick up.

Worth rebasing rather than merging: a few of the files you touched were adjusted in that review pass, so a rebase will surface those as conflicts you can resolve deliberately, where a merge would quietly take the older versions.

One heads-up for when you do, there are some debug leftovers still on the branch: `options.debug = true` in build.gradle, a `printStackTrace` callback added to `OpenPGPMessageProcessor`, and `System.out.println` calls in the Bc/Jce YubiKey factories that hex-dump recovered session keys and shared secrets.

Thank you.

vanitasvitae:
Hey! I rebased the PR and added some more tests and further progress. I like the changes you applied!
Though in the future I'd prefer to have git commit authorship preserved if possible :)

dghgit:
Merged with some revisions - I moved the external public key code under a new package org.bouncycastle.openpgp.api.operator. have a feeling it may serve us well to preserve the bcpg->openpgp->openpgp.api pattern. One thing, with debug, you can now just use:

./gradlew -Prelease.debug=true clean build

to do a debug build without having to change the build.gradle file.

The new API is now available at https://www.bouncycastle.org/betas as well

PR #2375, opened 2026-07-29
Add PGPKeyPairGenerator factory methods for OpenPGP brainpool keys

vanitasvitae:
This PR adds factory methods for generating OpenPGP ECDH and ECDSA keys to `PGPKeyPairGenerator`, as well as tests for each factory method.


ligefeiBouncycastle:
Thank you for this PR. It's now up on https://www.bouncycastle.org/betas if you want to try the revision.

PR #2383, opened 2026-08-01
guard ecdh session key length in the jce pgp decryptor

rootvector2:
the jce ecdh path derives `pLen` from `enc[0]`/`enc[1]` before its `checkRange` and outside the `try` that reports `PGPException`, so a truncated ECDH encrypted session key leaves `recoverSessionData` as an `ArrayIndexOutOfBoundsException`; spotted checking `0e75fc4` against the twin it fixed, whose release note records the jce side as already covered.


ligefeiBouncycastle:
Thank you for this PR. Several days ago I found the same issue in bc part and fix it. It's now up on https://www.bouncycastle.org/betas if you want to try the revision.

rootvector2:
thanks for taking this — I see it landed on main as `2664618`, will give the beta a try.

PR #2385, opened 2026-08-03
Fix engineGetCertificateAlias returning the alias of an unrelated certificate

### LINKED ISSUE #2384: PKCS12KeyStoreSpi.engineGetCertificateAlias returns the alias of an unrelated certificate (since 1.80)
### Summary

`KeyStore.getCertificateAlias(cert)` on a BC PKCS12 keystore returns the alias of a **different**
certificate. The returned alias exists in the store, so callers get a plausible but wrong answer with
no exception.

Present in 1.80, 1.80.2, 1.81.1, 1.83, 1.84 and 1.85 (all tested). 1.79 is correct.

### Cause

`PKCS12KeyStoreSpi.engineGetCertificateAlias` pairs two enumerations positionally:

```java
Enumeration c = certs.elements();
Enumeration k = certs.keys();

while (c.hasMoreElements())
{
    Certificate tc = (Certificate)c.nextElement();
    String ta = (String)k.nextElement();

    if (tc.equals(cert))
    {
        return ta;
    }
}
```

That holds only while `keys()` and `elements()` walk the same table in the same order. Commit
a5be993 ("Changed IgnoresCaseHashtable.keys() to copy the table content to ensure consistency",
first released in 1.80) changed `keys()` to enumerate a copy, while `elements()` still enumerates the
original:

```java
public Enumeration keys()
{
    return new Hashtable(orig).keys();   // copy
}

public Enumeration elements()
{
    return orig.elements();              // original
}
```

`new Hashtable(Map)` sizes its bucket array as `max(2 * size, 11)`, whereas `orig` reached its
capacity by incremental rehashing. Once the store holds enough entries for those to differ, the two
enumerations no longer agree and `ta` is the key at position *i* of the copy while `tc` is the value
at position *i* of the original.

`engineGetCertificateAlias` was not touched by that commit and is unchanged from `r1rv79` through
`r1rv85`, so this looks like an unintended side effect rather than a deliberate change.

### Reproduction

```java
Security.addProvider(new BouncyCastleProvider());

KeyStore store = KeyStore.getInstance("PKCS12", "BC");
store.load(null, null);

// 12 entries is enough for the original table and a copy of it to differ in layout
for (int i = 0; i != 12; i++)
{
    store.setCertificateEntry("cert-" + i, certs[i]);
}

for (int i = 0; i != 12; i++)
{
    System.out.println("cert-" + i + " -> " + store.getCertificateAlias(certs[i]));
}
```

Expected `cert-i -> cert-i` for every entry. On 1.80+ the mapping is permuted.

Originally hit on a real store holding one `PrivateKeyEntry` with a three-certificate chain plus four
trusted certificate entries, where every chain certificate resolved to an unrelated alias:

| chain certificate | 1.79 | 1.80 .. 1.85 |
|---|---|---|
| `CN=level2` (leaf) | `key-entry` | `trusted-c` |
| `CN=level1` | `CN=level1` | `trusted-b` |
| `CN=root` | `CN=root` | `key-entry` |

Every certificate's fingerprint and the full alias list are identical across versions — only the
lookup result changes. A store small enough that the table has not yet grown will not reproduce it,
which is likely why it went unnoticed.

### Impact

Silent misidentification in any code that maps a certificate back to its keystore alias — certificate
chain classification, trust-store bookkeeping, deciding whether a received certificate is already
stored. Nothing throws, so it fails quietly. In the application where I found it, an incoming
certificate was reported as already present (and so not stored) and then associated with a different
peer's certificate.

### Suggested fix

The copy in `keys()` was added deliberately, so the caller is the better place to fix: look the value
up by key instead of relying on the two enumerations agreeing.

```java
for (Enumeration k = certs.keys(); k.hasMoreElements();)
{
    String ta = (String)k.nextElement();

    if (cert.equals(certs.get(ta)))
    {
        return ta;
    }
}
```

The same applies to the `keyCerts` loop below it. PR to follow.

Note the `prov/src/main/jdk1.3` and `prov/src/main/jdk1.4` copies of this class still have
`return orig.keys();`, so they behave correctly today, but they carry the same order-dependent
pairing in `engineGetCertificateAlias`.

### Environment

- bcprov-jdk18on 1.80 – 1.85 (1.79 unaffected)
- JDK 21 and JDK 25
- `KeyStore.getInstance("PKCS12", "BC")`

rimuln:
Fixes #2384.

`PKCS12KeyStoreSpi.engineGetCertificateAlias` walked `certs.elements()` and `certs.keys()` in
lockstep, taking the certificate from one enumeration and the alias from the other. That relies on
both enumerations visiting the table in the same order, which stopped being true in 1.80: a5be993
changed `IgnoresCaseHashtable.keys()` to enumerate `new Hashtable(orig)` while `elements()` still
enumerates `orig`. `new Hashtable(Map)` sizes its bucket array from the entry count, whereas `orig`
reached its capacity by incremental rehashing, so once a store holds enough entries the two layouts
differ and the method returns the alias of an unrelated certificate — silently, with no exception.

Since the copy in `keys()` was introduced deliberately, this fixes the caller instead of reverting
it: the value is now looked up by key, so the result no longer depends on two enumerations agreeing.
The same change is applied to the `keyCerts` loop below it — `keyCerts` is a plain `Hashtable` whose
`keys()` and `elements()` do agree, so it was not broken, but the pairing was the same latent
fragility and removing it costs nothing.

`testGetCertificateAlias` in `PKCS12StoreTest` covers it: 12 certificate entries, then every
certificate is looked up and must return its own alias, plus a negative case for a certificate that
is not in the store. Twelve is chosen so the backing table has grown past its initial capacity — with
only a handful of entries the original and the copy can share a layout and the defect does not
reproduce, which is probably why it went unnoticed. The test fails before this change and passes
after.

Not addressed here: the `prov/src/main/jdk1.3` and `prov/src/main/jdk1.4` copies of this class still
have `return orig.keys();` and so behave correctly today, but their `engineGetCertificateAlias`
carries the same order-dependent pairing. Happy to extend the PR to those trees if you would prefer
them hardened too.

Prepared with AI assistance (noted in the commit trailer), offered under the Bouncy Castle License
per CONTRIBUTING.md.

### DIFF
--- a/prov/src/main/java/org/bouncycastle/jcajce/provider/keystore/pkcs12/PKCS12KeyStoreSpi.java
+++ b/prov/src/main/java/org/bouncycastle/jcajce/provider/keystore/pkcs12/PKCS12KeyStoreSpi.java
@@ -438,29 +438,24 @@ public class PKCS12KeyStoreSpi
     public String engineGetCertificateAlias(
         Certificate cert)
     {
-        Enumeration c = certs.elements();
-        Enumeration k = certs.keys();
-
-        while (c.hasMoreElements())
+        // look the certificate up by alias rather than advancing keys() and elements() in
+        // lockstep - IgnoresCaseHashtable.keys() enumerates a copy of the table while
+        // elements() enumerates the original, so the two orders are not required to agree.
+        for (Enumeration k = certs.keys(); k.hasMoreElements();)
         {
-            Certificate tc = (Certificate)c.nextElement();
             String ta = (String)k.nextElement();
 
-            if (tc.equals(cert))
+            if (cert.equals(certs.get(ta)))
             {
                 return ta;
             }
         }
 
-        c = keyCerts.elements();
-        k = keyCerts.keys();
-
-        while (c.hasMoreElements())
+        for (Enumeration k = keyCerts.keys(); k.hasMoreElements();)
         {
-            Certificate tc = (Certificate)c.nextElement();
             String ta = (String)k.nextElement();
 
-            if (tc.equals(cert))
+            if (cert.equals(keyCerts.get(ta)))
             {
                 return ta;
             }
--- a/prov/src/test/java/org/bouncycastle/jce/provider/test/PKCS12StoreTest.java
+++ b/prov/src/test/java/org/bouncycastle/jce/provider/test/PKCS12StoreTest.java
@@ -2417,6 +2417,46 @@ public class PKCS12StoreTest
         keyStore.getEntry("cycle", new KeyStore.PasswordProtection("test".toCharArray()));
     }
 
+    private void testGetCertificateAlias()
+        throws Exception
+    {
+        KeyStore keyStore = KeyStore.getInstance("PKCS12", BC);
+
+        keyStore.load(null, null);
+
+        // enough entries that the backing table has grown past its initial capacity: a copy of
+        // the table is sized from its entry count instead, so the two end up with different
+        // bucket layouts and any order-dependent pairing of keys() with elements() diverges.
+        KeyPair keyPair = TestUtils.generateRSAKeyPair();
+        int certCount = 12;
+
+        X509Certificate[] certs = new X509Certificate[certCount];
+        String[] aliases = new String[certCount];
+
+        for (int i = 0; i != certCount; i++)
+        {
+            aliases[i] = "cert-" + i;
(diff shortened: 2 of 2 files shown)


dghgit:
Thanks, merged with minor revision, now up on https://www.bouncycastle.org/betas

PR #2371, opened 2026-07-22
Update SMIMETest.java

cragkhit:
Hi. We are researchers from Mahidol University, Thailand, and the State University of Ceará, Brazil, working on a research project for improving open-source projects by using the latest accepted answer from Stack Overflow that matched your code snippet. We found this recommendation for improving your code from https://stackoverflow.com/a/40056844.

Note: Our study is approved by the Institutional Review Board of Mahidol University. You can find the participant information sheet explaining this study https://drive.google.com/file/d/1ml5AqrtWQ9pnifTQyTFTcWQmwp6RuPA7/view?usp=sharing.

Disclosure: This contribution was prepared with assistance from OpenAI Codex. I reviewed and verified the resulting change.

----

## Proposed change

Added identity and null checks before comparing the byte arrays.

## Testing

`./gradlew :util:compileTestJava :util:test`

Result: Passed

PR #2386, opened 2026-08-03
Update SMIMETest.java

cragkhit:
Hi. We are researchers from Mahidol University, Thailand, and the State University of Ceará, Brazil, working on a research project for improving open-source projects by using the latest accepted answer from Stack Overflow that matched your code snippet. We found this recommendation for improving your code from https://stackoverflow.com/a/40056844.

Note: Our study is approved by the Institutional Review Board of Mahidol University. You can find the participant information sheet explaining this study at https://drive.google.com/file/d/1ml5AqrtWQ9pnifTQyTFTcWQmwp6RuPA7/view?usp=sharing.

Disclosure: This contribution was prepared with assistance from OpenAI Codex. I reviewed and verified the resulting change.

This replaces #2371, whose branch contained unrelated divergent history.

## Proposed change

Add identity and null checks before comparing the byte arrays.

## Testing

- `./gradlew :util:compileTestJava` passed.
- `./gradlew :util:test` compiled and ran, but 10 unrelated C509CertificateTest cases failed with FileNotFoundException because their resource files were unavailable in the checkout.
- The repository Gradle configuration does not expose SMIMETest as an individually selectable test.

### DIFF
--- a/util/src/test/java/org/bouncycastle/asn1/smime/test/SMIMETest.java
+++ b/util/src/test/java/org/bouncycastle/asn1/smime/test/SMIMETest.java
@@ -26,6 +26,14 @@ public class SMIMETest
         byte[]  a,
         byte[]  b)
     {
+        if (a == b)
+        {
+            return true;
+        }
+        if (null == a || null == b)
+        {
+            return false;
+        }
         if (a.length != b.length)
         {
             return false;


dghgit:
Appreciate the effort to improve the library. It's an interesting idea, although in this case it's really the wrong spot - it's a test class and the method should never be called with anything that's null. Perhaps more importantly, the patches "if (a == b)" will return true if both a and b are null, which would be a disaster for the test's effectiveness.

For these reasons, this one I will not merge, however I would be interested to see what else you come up with.

PR #2401, opened 2026-08-20
check signed-message length in AIMerSigner.verifySignature

rootvector2:
`AIMerSigner.verifySignature` slices the signature out of the `message || signature` envelope at offset `message.length` without checking the buffer is that long, so a truncated signature throws `ArrayIndexOutOfBoundsException` out of `Signature.verify()` and a valid one with bytes appended still verifies (found by feeding wrong-length signatures to every BCPQC `Signature` service); it now requires exactly `message.length + getSignatureBytes()` and returns false otherwise, the guard the Falcon, Faest, Mayo, Snova and QRUOV signers already carry.


dghgit:
Thanks for the patch, now merged and up on https://downloads.bouncycastle.org/betas - one minor change, we optimised the test. Things are starting to take a while...

PR #2406, opened 2026-08-24
use constant-time comparison for srp-6a evidence messages

rootvector2:
`SRP6Server.verifyClientEvidenceMessage` and `SRP6Client.verifyServerEvidenceMessage` check the peer's evidence message (`M1`/`M2`, a keyed authenticator derived from the shared secret `S`) with `BigInteger.equals`, which walks the magnitude words and returns at the first difference, so the compare time depends on how many leading bytes match a value the unauthenticated peer is trying to guess. Switched both to `Arrays.constantTimeAreEqual` over the fixed encodings, matching the sibling `JPAKEUtil` MacTag check, in the lightweight `crypto.agreement.srp` package and the BCTLS `tls.crypto.impl.jcajce.srp` copies. Found while auditing the SRP path after the recent `createBlindedExponent` modPow hardening in these same classes; the accept/reject result is unchanged and `SRP6Test` now covers the evidence-verification path.


dghgit:
Thanks for the PR. Now merged and up on https://www.bouncycastle.org/betas code should sync up here shortly.

PR #2409, opened 2026-08-26
Allow Kangaroo doFinal after doOutput

tonywasher:
The Kangaroo doFinal call performs a check to ensure that it is not already squeezing, and throws an exception if it is already squeezing. However multiple doOutput calls followed by a doFinal is a valid sequence and is supported on all other Xof models (SHAKE/Blake3/Ascon). 

The check is redundant and should be deleted.


dghgit:
Thanks for the PR, now merged with revisions. Update is on https://www.bouncycastle.org/betas - source changes should appear on github shortly.

PR #2415, opened 2026-08-30
surface malformed TSTInfo as TSPException in TimeStampToken

rootvector2:
`TimeStampToken` parses the attacker-controlled `TSTInfo` eContent inside a try that catches only `CMSException`, so a well-formed `TimeStampResp` carrying a malformed `TSTInfo` (empty/short `SEQUENCE`, a non-`SEQUENCE`, or truncated DER) lets an unchecked `IllegalArgumentException`/`ClassCastException`/`NoSuchElementException` escape the `TimeStampResponse(byte[])` / `(InputStream)` path's declared `throws TSPException, IOException` — the sibling `TimeStampResponse(DLSequence)` ctor already wraps `new TimeStampToken(...)`, so this rethrows the parse failure as `TSPException` to match it.


dghgit:
Thanks for the patch, this is now merged and available at https://downloads.bouncycastle.org/betas

PR #2430, opened 2026-09-08
OpenPGP Smart Card Signature Support

vanitasvitae:
This PR extends the OpenPGP Smart Card implementation with signature support.

It applies the following changes:
* Bump `yubikit` dependency to `3.2.0`
* Detect and properly handle external keys in `AbstractOpenPGPDocumentSignatureGenerator`
* Allow the user to set a custom `PGPContentSignerBuilderProviderFactory` for hardware-based signature keys
* Reduce `PublicKeyCryptoCallback` method set to two methods (one for RSA/ElGamal, one for `PublicKey`-based handshake algorithms ECDH, X25519, X448).
* Add helper classes to convert between JcaJce and BC key types
* Add `ExternalOpenPGPKeyUtils` to convert `OpenPGPKey`/`OpenPGPCertificate` objects to external `OpenPGPKey` objects
* Move low-level `sign` and `decrypt` methods to `OpenPGPSmartCard` to allow for low-level emulation of hardware devices using simulated software keys


pg/src/test/java/org/bouncycastle/openpgp/test/JceExternalPublicKeyDataDecryptorFactoryBuilderTest.java
```diff
@@ -96,14 +102,28 @@
...
                 {
+                    Cipher c;
                     try
                     {
-                        Cipher c = Cipher.getInstance("RSA/ECB/PKCS1Padding", "BC");
+                        switch (keyAlgorithm)
+                        {
+                            case PublicKeyAlgorithmTags.RSA_GENERAL:
+                            case PublicKeyAlgorithmTags.RSA_ENCRYPT:
+                                c = Cipher.getInstance("RSA/ECB/PKCS1Padding", "BC");
```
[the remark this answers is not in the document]
vanitasvitae: Copied over from https://github.com/bcgit/bc-java/blob/main/pg/src/test/java/org/bouncycastle/openpgp/test/JceExternalPublicKeyDataDecryptorFactoryBuilderTest.java#L104
I don't think there is much we can do about it, since the padding is mandated by RFC2440, RFC4880, RFC9580.

pgsc/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/ec/BCECPublicKeyHelper.java
vanitasvitae, line 1: This class was necessary to cleanly convert BC keys to JCAJCE keys due to package-only visible constructors.
I'm currently running into some issues with JPMS complaining about the package being present in multiple modules, so this apparently is not a good solution either.
Can we have public constructors instead?

pgsc/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/edec/EDECPublicKeyConverter.java
vanitasvitae, line 1: The same issue as `BCECPublicKeyHelper` applies here as well.

vanitasvitae:
I addressed some of the points from the private report.

Some remarks:
I don't necessarily agree with items 5 and 8:
* Finding 5: I think hard-coding rejection of individual algorithms is not the right approach. Algorithm support may change in the future (although support for ElGamal is pretty unlikely). I think a better approach would be to detect, which algorithms the card currently uses and to compare the keyAlgorithm octet that way. I'll address this in a future patch with proper algorithm detection support.
* Finding 8: I'm a fan of addressing legacy breakage ASAP. Keeping read support for legacy identifiers of an outdated format may hide the fact that an implementation used an outdated scheme to encode fingerprints, which increases both code complexity as well as the potential of that outdated behavior to remain unfixed. Therefore I'm in favor of not adding read support for the outdated scheme.

dghgit:
I've rolled this one in with a few minor revisions. The current version is up on https://www.bouncycastle.org/betas

PR #2431, opened 2026-09-08
reject pbmac1 keylength below 20 octets in pkcs#12

rootvector2:
found auditing the RFC 9579 PBMAC1 verify path: `validateKeyLength` in both PKCS#12 `PKCS12Util` helpers (`pkcs.util` and `jcajce.provider.keystore.pkcs12`) bounded the attacker-supplied PBKDF2 `keyLength` only from above, so a PFX declaring `keyLength` of one octet verified its integrity MAC against a one-byte HMAC key that is trivially brute-forced, and both now reject a `keyLength` below 20 octets before deriving, per RFC 9579 sec. 9 (matching OpenSSL's `keylen < 20` check) and the `20-64 bytes` floor their own comment already claimed.

AI tooling was used to help prepare this change.


dghgit:
Thanks for the patch, now merged with some revisions and available in the 1.86 release candidate on https://downloads.bouncycastle.org/betas

PR #2437, opened 2026-09-11
Add CMS SignedData and PKIX CertPath tests for the Composite ML-DSA parameter sets

Arpan0995:
Composite ML-DSA (draft-ietf-lamps-pq-composite-sigs) is wired through DefaultCMSSignatureAlgorithmNameGenerator and the DefaultSignatureAlgorithmIdentifierFinder used on the CMS and cert-path paths for all eighteen IANA parameter sets. Plain JCA sign/verify for these sets is covered by CompositeSignaturesTest, and the Composite ML-KEM EnvelopedData path has CompositeKEMEnvelopedDataTest, but there was no CMS SignedData or PKIX CertPath (chain) coverage on the Composite ML-DSA signature side.

This adds CompositeMLDSASignedDataTest with two cases over all eighteen parameter sets:

- a CMS SignedData sign and verify round-trip through SignerInformation.verify, asserting the SignerInfo signature-algorithm OID is the composite OID, with a negative control that the signature does not verify against a different composite key; and
- a CA to end-entity chain validated through CertPathValidator("PKIX"), with a negative control that an end-entity certificate signed by a key outside the trust anchor is rejected.

Test only, no production change.


dghgit:
Thanks for the patch. Now up on https://www.bouncycastle.org/betas

PR #2442, opened 2026-09-12
DTLS 1.3 post-handshake: KeyUpdate, NewSessionTicket and ACKs (RFC 9147), part 4 of 4; closes #1468

### LINKED ISSUE #1468: DTLS 1.3 support

mondain:
DTLS 1.3 (RFC 9147), part 4 of 4: post-handshake. Closes #1468.

Stacked on #2441 (client and server handshakes), which stacks on #2440 (reliable handshake) and #2439
(record layer). Review those first; this PR's own diff is against #2441.

This adds key update in both roles, NewSessionTicket handling, and acknowledgements at the application
epochs. With it, the four parts together give bctls a working DTLS 1.3 implementation sufficient for
WebRTC: a certificate-authenticated 1-RTT handshake with HelloRetryRequest and optional client
authentication, ACK-driven recovery from loss, exporters, and DTLS-SRTP over DTLS 1.3.

## What a maintainer will want to check first

**Why four files shared with TLS over TCP are touched.** This is the first question the diff invites, and
the answer is not obvious: `TlsAEADCipher.rekeyDecoder()` already resolves through the connection security
parameters, so bctls *already has* a post-handshake-safe rekey path that would need none of these changes.

It cannot be used here. `rekey*` mutates the cipher in place, and DTLS needs the pre-update cipher to stay
usable in both directions. On the read side, RFC 9147 section 8 requires retaining the pre-update keys until
something decrypts under the new ones. On the write side, records keep going out at the old epoch for the
whole unacknowledged-KeyUpdate window. In-place rekey destroys exactly the material both depend on. Building
a new cipher is therefore forced, and once a cipher is built after `handshakeComplete` the three constructor
sites must be able to resolve security parameters — hence `TlsCryptoParameters.getSecurityParameters()`,
handshake-else-connection, mirroring the accessor that already exists on `TlsContext`.

**Why it is neutral.** `AbstractTlsContext.handshakeComplete` *moves* the reference: one
`SecurityParameters` instance, assigned to the connection field and nulled from the handshake field. So
during a handshake the new accessor and `getSecurityParametersHandshake()` return the same object, and every
existing caller of the changed sites runs inside the handshake window, renegotiation included. TLS 1.3 over
TCP never reaches those constructors post-handshake at all, because it rekeys in place. The only reachable
difference is a widening: a call that used to throw now returns a cipher.

**The one behavioural change to DTLS 1.2**, from part 3 and restated here because it is easy to miss:
`notifySecureRenegotiation` fires from `generateServerHello` after `processClientExtensions`, which is where
`TlsServerProtocol` has always called it. No wire bytes move.

## The two rules of RFC 9147 section 8

They have different triggers, and conflating them is the natural error:

- The **sender** must not transmit at the new epoch until its KeyUpdate is **acknowledged**.
- The **receiver** must retain the pre-update keys until the **first successful decryption** at the new epoch.

An acknowledgement is not a decryption. The two are implemented in separate commits and the separation is
structural: the only call that installs a pending write epoch is in the ACK path, and the only call that
releases a retained read epoch is in the decrypt path.

Supporting this needed two changes to the record layer. The read side previously resolved an incoming
record against two candidates; under key update a third epoch is live, and a record arriving under it would
have matched neither and been dropped. It now resolves through one ordered, bounded, newest-first set, which
makes section 4.2.2's "most recent past epoch with matching bits" fall out of the structure rather than
being special-cased. The write side gained an epoch that can be derived and held without being installed,
because the existing pending-epoch machinery commits only when both directions advance together, which is
exactly what key update does not do.

## A deliberate deviation from a MUST

RFC 8446 4.6.3 says a peer receiving `update_requested` must send its own KeyUpdate before its next
application data record. RFC 9147 5.8.4 says it must not send one while another is unacknowledged. When our
own KeyUpdate is outstanding and an `update_requested` arrives, both cannot be honoured without stopping
application data indefinitely.

This follows RFC 9147, on three grounds. It is the DTLS specification constraining a TLS rule it inherits.
Its rule is load-bearing: with two epoch bits on the wire, a second unacknowledged KeyUpdate makes epoch
reconstruction ambiguous, whereas 8446's rule is timeliness hygiene. And the third option, blocking
application data until the outstanding update clears, would be a self-inflicted stall on the data path in
response to a peer-controlled message.

The obligation is recorded and discharged on the first send after the acknowledgement, so it is deferred
rather than dropped. It is stated at the site and exercised by a test whose mutation is "conform to RFC 8446
instead". Flagging it explicitly because a reader who finds only the 8446 quotation would reasonably read
this as an oversight.

## Test seams

Four package-private methods on main sources exist only for tests, one of them on the public class
`DTLSTransport`. None widens the published API. Three exist because the conditions they reach — a 2^20
sequence number, a resend interval, twice the MSL — take minutes of wall clock to reach naturally.

A public `requestKeyUpdate()` on `DTLSTransport` would remove one of the four and was deliberately not
added: it would be permanent public API, designed under review pressure, with no second implementation to
check the semantics against, and `TlsProtocol` has no TLS equivalent. Happy to add it if you would prefer
that shape — it seems better decided by you than by us.

## Known gaps

- **No interoperability testing against any non-Bouncy Castle implementation, for any part of the series.**
  This is the honest headline risk. Two facts are covered by byte-level tests against independently computed
  values: the DTLS 1.3 transcript header form, and the HelloRetryRequest synthetic transcript. Everything
  else is currently proven BC-to-BC, which cannot catch an error both peers make identically.
- The automatic key update threshold cannot be driven end to end. Jumping the sequence number far enough to
  reach it breaks RFC 9147 4.2.2 epoch reconstruction, so every later record fails to decrypt. The
  arithmetic is unit-tested; its effect on a live connection is not.
- Only one key update per direction is driven end to end. A second-update defect fixed in this PR, where
  acknowledgements from a peer still writing at the older epoch were discarded, is covered by unit tests.
- Exporter output could not be re-checked after the handshake, because bctls clears the exporter master
  secret at handshake return. The shipped test asserts that refusal, with a tripwire. A probe with the
  secret retained showed byte-identical material after a key update in both directions, so the answer is
  known even though it cannot be asserted through the shipped API.
- No literal zeroisation of superseded key material. The code drops the last reference, which is what the
  handshake path already does; `TlsCipher` has no destroy operation and adding one is its own change.
- Still out of scope across the series: PSK and resumption, 0-RTT early data, connection IDs (RFC 9146),
  post-handshake client authentication, a cookie-only HelloRetryRequest, and a stateless DTLS 1.3 front end
  analogous to `DTLSVerifier`.

## Tests

892 in the `tls` module, no failures; checkstyle clean; `tls/src/main/java` within the legacy Java 1.4/1.5
Ant build constraints.

Every load-bearing assertion in this PR has a reported mutation that kills it. The ones worth naming:

- The no-send-before-acknowledgement gate survives having every field and record-number assertion stripped
  out, and still fails on the epoch observed on the wire.
- A mutation that builds the new epoch's cipher without updating the traffic secret fails an assertion that
  the new epoch cannot read what the pre-update keys produced. This is the first test in the series that
  would catch a derivation silently returning the same key material; earlier epoch tests key both sides
  identically on purpose, to isolate the numbering.
- Reversing the epoch resolution order fails a test built so that ordering is the only thing that can
  decide the outcome: both candidates alias on the low bits *and* carry identically-keyed ciphers, so either
  could decrypt, and the assertion is on which one consumed the record.

## Disclosure

This work was produced with generative-AI assistance, per the contributing guidelines. Every design
decision, every security finding and its fix, and every test were reviewed; the RFC citations were checked
against the specification text rather than recalled.


mondain:
This closes out my series to use BC in our WebRTC server; I've built it locally as 1.86 and tested with my server updates and so far I've found no issues while testing DTLS 1.3.

mondain:
Any moderators available for reviews? @peterdettman @dghgit

mondain:
@JonathanLennox, an update on the two items you offered to add.

message_seq wraparound (#304). Added to the series. message_seq is a uint16 and writeUint16 truncated silently, so the 65537th message would have reused an earlier sequence number. There were two send-side counters, so two guards, both throwing internal_error before anything is written:
  - DTLSReliableHandshake.sendMessage (part 2, #2440)
  - DTLS13PostHandshake.sendKeyUpdate (part 4, this PR). The check runs before the next write epoch is derived, so a refusal leaves the keys and write epoch untouched.

Tests cover the last valid sequence number (0xFFFF) still going out and 0x10000 being refused. I confirmed each wrap test fails with its guard removed. The full :tls:test run passes.

The post-KeyUpdate unexpected_message check (#289) is not done yet.

Parts 2 to 4 were rebased to carry this, so their commit hashes changed: part 3 is now d74876735c and part 4 is c2790c37c9. Anything you have built on the old part 4 tip (b274dc3402) will need to be rebased or re-fetched.

PR #2446, opened 2026-09-14
Unwrap the RFC 9709 content-encryption algorithm at every CMS recipient key-size and policy check

Arpan0995:
Under RFC 9709 the `EncryptedContentInfo` names `id-alg-cek-hkdf-sha256` and carries the content-encryption `AlgorithmIdentifier` in its parameters. The 1.86 fix for CVE-2026-71892 resolves it, per its release note: "The recipient now dispatches on the content-encryption algorithm OID". The sibling call sites do not. `JceKEKRecipient`, `JceKTSKeyTransRecipient`, `JceKEMRecipient` and `JceKeyTransRecipient`'s own `id-ori-kem` branch pass the outer identifier to `EnvelopedDataHelper.keySizeCheck`, which registers no key size for it, so `setKeySizeValidation(true)` checks nothing. The 1.85 policy controls read the outer identifier on every recipient family, key transport included: an allowed set naming `aes256-CBC` refuses the message outright, one naming `id-alg-cek-hkdf-sha256` admits any content algorithm, and `setMinimumTagSize` never sees a tag size.

Reproduced on the released 1.86 jars and on classes built from `origin/main` (`ab16374d37`), unchanged in `org.bouncycastle.cms` since `r1rv86`. A 32-byte CEK under an advertised `aes128-CBC` draws `CMSException: Expected key size for algorithm OID not found in recipient.` from all four with HKDF off; with it on, key transport still refuses and KEK, KTS and ML-KEM return the content. AES-128-GCM with a 96-bit tag under `setMinimumTagSize(128)` raises `CMSTagLengthException` with HKDF off and recovers with it on for all four, key transport included: the tag floor was not in the 1.86 fix.

This change:

- hoists the `id-alg-cek-hkdf-sha256` unwrap into `EnvelopedDataHelper.keySizeCheck`, so all five key-size sites validate against the carried algorithm and the special case in `JceKeyTransRecipient.extractSecretKey` collapses to the plain call (the `id-ori-kem` branch carries no test, since no shipped generator emits such a `KeyTransRecipientInfo`);
- adds `AbstractRecipient.checkContentAlgorithm`, which resolves the wrapper once and then applies the allowed set and the tag floor, replacing the identical prologue in the nine recipients carrying it (shape only in the other five: only the four `Jce` recipients that derive the CEK reach an RFC 9709 message today); and
- adds five cases to `NewEnvelopedDataTest`: the KEK, KTS and ML-KEM key-size sites, and the allow-list and tag-floor halves on `JceKeyTransRecipient`. They are the only failures against the unchanged sources; with the change all pass, as do the other 224 tests in the 15 CMS enveloped-, auth-enveloped- and authenticated-data classes.

One behaviour change: a caller who listed `id-alg-cek-hkdf-sha256` in an allowed set must now list the content algorithms themselves.

Base tree only, as the 1.86 fix was; the legacy `pkix/src/main/jdk1.1` overlays are left alone, and no `module-info` / OSGi change is needed. A release-note entry is in the 1.87 defects block; happy to reword it.


dghgit:

PR #2455, opened 2026-09-19
Reuse a modular power when initializing generic sqrt_ratio constants

carstenartur:
Since c2 = 2*c3 + 1, derive c7 and c6 from z^c3 instead of computing two independent powers (RFC 9380 F.2.1.1). Preserve the exact constants and leave the per-call sqrtRatio algorithm unchanged.

Cover direct-power equality, zero exponents and negative/unreduced bases, repeated root equations, and the reduced exponentiation count.

AI-assisted implementation; benchmark infrastructure is kept in the research fork rather than included in this upstream candidate.

Reuse a modular power when initializing generic sqrt_ratio constants

## Change

Avoid two independent modular powers in the `GenericSqrtRatioCalculator` constructor. Since `c2 = 2*c3 + 1`, compute:

```text
t  = z^c3 mod q
c7 = t*z mod q
c6 = t*c7 mod q
```

These are exactly the constants from RFC 9380 F.2.1.1. One modular exponentiation replaces two, with two modular multiplications. The identity includes exponent zero and negative/unreduced bases without inversion. The per-call `sqrtRatio` method is unchanged. No API, cache, randomness, security-parameter or dependency changes are introduced.

This is one commit on `ab16374d37c7e18c4090eb8838ebbd72a92593f2`, with only three production/test files. Research workflows and benchmarks stay in the fork.

## Public API measurements

The real `HashToEllipticCurve.getInstance(profile, dst).hashToCurve(message)` path was measured with 64 rotating deterministic 32-byte messages. The two libraries have identical class names and differ only in the calculator class.

| Profile | Less time, Java 21 | Less time, Java 25 |
|---|---:|---:|
| P-256 | 10.92% | 12.06% |
| P-384 | 15.09% | 16.53% |
| P-521 | 16.29% | 15.71% |
| Equal-profile geometric comparison | 14.13% | 14.79% |

All six primary JMH intervals are separated. Factory-only time is about 46% lower; all six reused-instance intervals overlap. This is a warm-JVM initialization/first-use improvement, NOT a steady-state, cold-start, all-curve or whole-application speedup.

JMH 1.37: 2 forks, 3 warmups and 5 measurements of 500ms, one thread, GC profiler; all cases and 180 primary samples per JVM retained. JVMs ran on separate hosted runners, so this is not a controlled comparison of Java versions.

[Full results and raw-artifact identifiers](https://github.com/carstenartur/bc-java/blob/7e6fec7c5c61f1a5dbfcb12680504489e8e5d038/research/hash2curve-public-api/RESULTS-2026-09-19.md) | [Fixed plan and runnable harness](https://github.com/carstenartur/bc-java/tree/b1d324423d4c83a05f1e1cda423bc5bd07594696/research/hash2curve-public-api)

## Validation

The complete default repository `test :core:checkstyleMain` run passed with 3,602 JUnit cases, zero failures/errors/skips, and zero core main-source Checkstyle findings. It exercised the same production and test blobs as this patch. A separate job checked out exact candidate `3d56837c3fe6c4a30fc7639b806958649e143210` and passed all 181 hash-to-curve cases and core main-source style. Artifact hashes, actual XML, zero exit receipts and terminal build success were verified: [full validation](https://github.com/carstenartur/bc-java/blob/baf2973b452b192c087468a742381eb2dcdaa808/research/hash2curve-public-api/UPSTREAM-VALIDATION.md).

Added tests compare actual constants against the original powers, cover edge cases, check repeated root equations/QR flags and the reduced exponentiation count. The count test fails on the original constructor; a deliberately corrupted constant is detected. Existing RFC-vector tests are unchanged. The API experiment also checks 201 exact points, fresh/reused equivalence and reference outputs before each trial.

## AI assistance

ChatGPT assisted with implementation, tests and benchmark preparation. The algebra, code and executable evidence were reviewed, and an additional automated Copilot review reported no findings. Upstream maintainer review is requested. No increased-security, constant-time BigInteger or autonomous mathematical-discovery claim is made.


peterdettman:
Rebased and merged, thanks for the PR. Note that one test was dropped as redundant and pinning an implementation detail.

